PR #1, opened 2019-08-09
Feature/sign up

t-romani:
add devise config
create user model, migrate users (devise) table
create registration controller
sign_up successfully creates new user

some added gems: annotate, figaro, jbuilder, pg,


gonzaM (approved):
Good job!
For future PRs try to separate the logic to avoid having big PRs. In this PR you include things for sign up, email confirmation, reset password.

Gemfile
```diff
@@ -36,10 +37,17 @@ gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
+  gem 'factory_bot_rails', '~> 5.0', '>= 5.0.2'
   gem 'rspec-rails', '~> 3.8', '>= 3.8.2'
 end
 
+group :test do
+  gem 'database_cleaner', '~> 1.7'
```
Alxzu: Do we still need to use this gem? @gonzaM

Alxzu: Check here https://anti-pattern.com/transactional-fixtures-in-rails

JuanfraM: good one! I didn't know. Thanks 💪

JuanfraM: @t-romani we can try to remove the gem and its configuration and test if it works

app/models/user.rb
```diff
@@ -0,0 +1,56 @@
...
+         :validatable, :confirmable, :trackable
+
+  enum gender: { male: 0, female: 1, other: 2 }
+
+  before_validation :init_uid
+
+  validates :email, presence: true, uniqueness: true
+  validates :full_name, presence: true
+  validates :gender, presence: true
+  validates :encrypted_password, presence: true
```
JuanfraM, line 47: this validate is not necessary since devise check it

app/models/user.rb
```diff
@@ -0,0 +1,56 @@
...
+  validates :uid, presence: true, uniqueness: { scope: [:provider] }
+  validates :provider, presence: true
```
JuanfraM, line 49: same for `uid` and `provider`

app/models/user.rb
```diff
@@ -0,0 +1,56 @@
...
+  include DeviseTokenAuth::Concerns::User
+
+  devise :database_authenticatable, :registerable, :recoverable, :rememberable,
+         :validatable, :confirmable, :trackable
+
+  enum gender: { male: 0, female: 1, other: 2 }
+
+  before_validation :init_uid
+
+  validates :email, presence: true, uniqueness: true
```
JuanfraM, line 44: when you're using the same validation for differents attributes you can write it on the same line:
`validates :full_name, :gender, presence: true`

spec/factories.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :user do
+    email { 'test_user@example.com' }
+    password { 'example' }
+    password_confirmation { 'example' }
```
JuanfraM, line 5: passwords are not necessary in the factory

spec/factories.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :user do
+    email { 'test_user@example.com' }
```
JuanfraM, line 3: you can use the gem [faker](https://github.com/faker-ruby/faker) to create fake data

spec/models/user_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+RSpec.describe User, :type => :model do
+  subject do
+    @user1 = FactoryBot.create(:user)
+  end
```
JuanfraM, line 5: I think this could be in one line:
subject { create(:user) }

spec/models/user_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  describe '#valid attributes' do
+    it 'is valid with valid attributes' do
+      subject
+      expect(@user1).to be_valid
```
JuanfraM, line 10: here you can set:
`expect(subject).to be_valid`

spec/models/user_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  describe '#unique mail, (uid,provider)' do
+    it 'has a unique email' do
+      subject
```
JuanfraM, line 16: why are u calling subject here  🤔 . Then you are using user2

spec/models/user_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      user2 = build(:user)
+      expect(user2).to_not be_valid
+    end
+
+    it 'has a unique (uid,provider)' do
+      subject
```
JuanfraM, line 22: same here and above

spec/request/user/create_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    it 'does get a sucessful answer' do
+      subject
+      expect(response).to be_successful
+    end
+
+    it 'does create a new user' do
+      expect { subject }.to change { User.count }.by(1)
+    end
+
+    it 'the attributes from params persist on db' do
```
JuanfraM, line 19: `it 'returns the user created'` sounds more appropriate

spec/request/user/create_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        user_params[:user][:full_name] = nil
+        subject
+        expect(response).not_to be_successful
+      end
+    end
+
+    context 'when invalid password' do
+      it '(different) it does not create a new user' do
+        user_params[:user][:password] = 'password'
+        user_params[:user][:password] = 'different'
```
JuanfraM, line 60: why twice 🤔

spec/request/user/create_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        subject
+        expect(response).to have_http_status(:unprocessable_entity)
+      end
+
+      it '(short) it does not create a new user' do
+        user_params[:user][:password] = 'short'
+        user_params[:user][:password] = 'short'
```
JuanfraM, line 67: same here, twice 🤔

spec/models/user_spec.rb
t-romani: this is done to test uniqueness of email, i need a user1 created in order to 'not being able to create' user2

spec/request/user/create_spec.rb
t-romani: it was supposed to be password & password_confirmation, my bad 😆

spec/factories.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
```
andresg4, line 1: I think this file should be called `users.rb` and located in the folder `spec/factories`. Then when you have a new model you would have a new factory file like `new_models.rb` in the same folder.

spec/request/user/create_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'rails_helper'
+
+RSpec.describe 'POST #create sign_up', type: :request do  
+  let!(:user_params) { { user: FactoryBot.attributes_for(:user) } }
+  subject do
```
andresg4, line 5: I think it would be better to leave an empty line after the `let`. Also, here in the specs you can use `attributes_for(:user)` without the `FactoryBot`.

spec/request/user/create_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      expect { subject }.to change { User.count }.by(1)
+    end
+
+    it 'the attributes from params persist on db' do
+      subject
+      expect(response.body).to include_json(
+        user: {
+          email: User.last.email,
+          gender: User.last.gender,
+          full_name: User.last.full_name
```
andresg4, line 25: Maybe in this case that you use `User.last` several times, you can define `let(:user) { User.last }` and use it here like:
```
email: user.email,
gender: user.gender,
full_name: user.full_name
```

spec/request/user/create_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      )
+    end
+  end
+
+  context 'when invalid' do
+    context 'when missing argument' do
+      it '(email) it does not create a new user' do
+        user_params[:user][:email] = nil
+        subject
+        expect(response).not_to be_successful
```
andresg4, line 36: Maybe here and in the tests below you could check for a more specific error status. And also you could check for the error message. Like:

```
context '(email) it does not create a new user' do
  before do
    user_params[:user][:email] = nil
    subject
  end

  it 'returns unauthorized status' do
    expect(response).to have_http_status(:unauthorized)
  end

  it 'returns error message' do
    expect(json[:error]).to eq('some error message')
  end
end
```

spec/models/user_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  describe '#valid attributes' do
+    it 'is valid with valid attributes' do
+      subject
+      expect(@user1).to be_valid
+    end
+  end
+
+  describe '#unique mail, (uid,provider)' do
```
andresg4, line 14: Maybe for this uniqueness and presence tests you could use the `shoulda-matchers` ([link](https://github.com/thoughtbot/shoulda-matchers)). With this gem you can check the presence and uniqueness validations easier and in one line. You can use: 

```
it { is_expected.to validate_uniqueness_of(:email) }
it { is_expected.to validate_presence_of(:email) }
```

For this to work correctly you have to add some config to the `rails_helper.rb` file ([link](https://github.com/thoughtbot/shoulda-matchers#rails-apps)).
Also [here](https://github.com/thoughtbot/shoulda-matchers#matchers) you have a list of the available matchers to use.

app/views/devise/mailer/confirmation_instructions.html.erb
```diff
@@ -0,0 +1,5 @@
+<p>Welcome <%= @email %>!</p>
```
andresg4, line 1: Maybe you can use `I18n` for the texts in the emails. You would have the text in `config/locales/en.yml`, and use them in the email views like `I18n.t('key_in_en.yml')`.
For example for this line you can have in the `en.yml`:
```
mailer:
  confirmation:
    title: "Welcome %{email}!"
```

And you can use it in the view like:
```
<%= t('mailer.confirmation.title', email: @email) %>
```

andresg4 (commented):
Nice work 👏 . I leave some comments.

app/views/devise/confirmations/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<h2>Resend confirmation instructions</h2>
```
Alxzu, line 1: Since this is an API I think you should remove this file, WDYT @JuanfraM @gonzaM ?

spec/request/user/create_spec.rb
t-romani: Where in particular should I define the User? 
Because due to the test, the user is created during the subject, then i should call the let!(:user) after the subject has been called so i can reference to the -last user- indeed, the thing is it wont allow me to call let inside the `it` block.

andresg4: You can define the `let` before the `context` starts. For example:
```
RSpec.describe 'POST #create sign_up', type: :request do  
  let!(:user_params) { { user: FactoryBot.attributes_for(:user) } }
  let(:user)         { User.last }

  subject do
    post '/api/v1/users', params: user_params, as: :json
  end
```
If you don't use the `!` (just `let`), the `User.last` isn't going to be called until you call `user`. So if you call `user` inside the `expect` the user will be already created.
You can see some explanation and examples [here](http://www.betterspecs.org/es/#let) and [here](https://mixandgo.com/learn/let-vs-instance-variables-in-rspec).

t-romani: Great, thanks!

app/views/devise/confirmations/new.html.erb
t-romani: Every one of those views was automatically created by devise, tbf I didnt even know I had them uploaded 😅

JuanfraM: @Alxzu agree, thanks!

JuanfraM (approved):

PR #2, opened 2019-08-15
Feature/sign in

t-romani:


spec/factories.rb
```diff
@@ -6,4 +6,8 @@
     full_name { 'Testereck Usernder' }
     gender { 2 }
   end
+
+  factory :confirmed_user, parent: :user do
```
jiv: I would rather go with a trait for this https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#traits

You can mix traits easily, it seems harder to mix factories

jiv (approved):

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Api
+  module V1
+    class SessionsController < DeviseTokenAuth::SessionsController
+      include ActAsApi
+    end
```
JuanfraM, line 13: I think you can do something similar as you did in the `RegistrationController`, you can override the method `create` and also `sign_in_params`, check at this:
[SessionController](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/sessions_controller.rb)

Then you can remove `devise_parameter_sanitizer.permit(:sign_in, keys: %i[email])` above in the `ApiController`

config/environments/development.rb
```diff
@@ -31,9 +31,23 @@
   config.active_storage.service = :local
 
   # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
+  # config.action_mailer.raise_delivery_errors = false
```
JuanfraM: remove this since you are setting it as true below

JuanfraM (commented):
Some comments/
Good job 👏 👏 👏

app/controllers/api/v1/confirmations_controller.rb
```diff
@@ -0,0 +1,8 @@
+module Api
+  module V1
+    class ConfirmationsController < DeviseTokenAuth::ConfirmationsController
+      include ActAsApi
+
```
gonzaM: Remove empty line

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'rails_helper'
+
+RSpec.describe 'POST #create sign_in', type: :request do
```
gonzaM: RSpec is not necessary

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  let!(:user_attrs) { { user: FactoryBot.attributes_for(:user) } }
+
+  subject do
+    post '/api/v1/users/sign_in', params: {
```
gonzaM: You can use something like: `new_user_session_path` check the routes, and use the name prefix with `_path`

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      email: user_attrs[:user][:email], password: user_attrs[:user][:password]
+    }, as: :json
+  end
+
+  context 'when valid' do
+    context 'when registered and confirmed' do
+      let!(:user) { FactoryBot.create(:confirmed_user) }
```
gonzaM: `FactoryBot` is not necessary, when you use the `create` here the tests will know that is FactoryBot

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      it 'does create a new session' do
+        expect { subject }.to(change { User.last.current_sign_in_at })
+      end
+    end
+  end
+
+  context 'when invalid' do
+    context 'when registered and not confirmed' do
+      let!(:user) { FactoryBot.create(:user) }
```
gonzaM: Same here and for all `FactoryBot`

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      it 'does not get a sucessful answer' do
+        post '/api/v1/users/sign_in', params: {
+          email: user_attrs[:user][:email], password: 'another'
+        }, as: :json
+        expect(response).not_to be_successful
+      end
+
+      it 'does not sign_in' do
+        post '/api/v1/users/sign_in', params: {
+          email: user_attrs[:user][:email], password: 'another'
```
gonzaM: Something you can do here:
In the subject above
```ruby
subject do
    post new_path, params: { email: user_attrs[:user][:email], password: password }, as: :json
end
```
Then you can redefine the `let` inside each context, and the subject will use that let.

```ruby
context 'when password not valid'
  let(:password) {  'invalid' }

  it .... do
    subject # will use subject with password = 'invalid'
    expect...
  end
end

context 'when password is a number'
  let(:password) {  123 }

  it .... do
    subject # will use subject with password = 123
    expect...
  end
end
```

gonzaM (approved):
Good job! Small comments

app/controllers/api/v1/sessions_controller.rb
t-romani: I dont fully understand the create method on the controller, time to come to work 😏

app/controllers/api/v1/api_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Api
+  module V1
+    class ApiController < ApplicationController::API
+      include DeviseTokenAuth::Concerns::SetUsedByToken
+
+      before_action :authenticate_user!
+      before_action :pepe
```
Alxzu: 😂

t-romani: 😆 shhhh, this never happened

app/controllers/api/v1/api_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Api
+  module V1
+    class ApiController < ApplicationController::API
```
JuanfraM: This controller should inherit from ApplicationController only not `::API` 🤔

app/controllers/api/v1/confirmations_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Api
+  module V1
+    class ConfirmationsController < DeviseTokenAuth::ConfirmationsController
+      include ActAsApi
```
JuanfraM: do you need this controller 🤔

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Api
+  module V1
+    class SessionsController < DeviseTokenAuth::SessionsController
+      include ActAsApi
+
+      def sign_in_params
+        params.require(:user).permit(:email, :password)
+      end
```
JuanfraM, line 8: nice!
Also, are you returning the user after sign in?

spec/factories/user.rb
```diff
@@ -0,0 +1,14 @@
...
+    email     { 'test_user@example.com' }
+    password  { Faker::Internet.password }
+    full_name { Faker::Name.name_with_middle }
+    gender    { 2 }
+
+    trait :confirmed do
+      after :create, &:confirm
+    end
+
+    factory :confirmed_user, traits: [:confirmed]
```
JuanfraM: you can remove this factory since you have created the trait above

spec/models/user_spec.rb
```diff
@@ -1,49 +1,17 @@
...
-    end
-
-    it 'is not valid without an gender' do
-      subject
-      user2 = build(:user, gender: nil)
-      expect(user2).to_not be_valid
+  describe 'email uniqueness' do
+    it 'should not be valid' do
+      user1 = create(:user)
+      expect(subject).to_not be_valid
```
JuanfraM: you can use the matcher `validate_uniqueness_of` and move this validation with the group above

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    context 'when registered and confirmed' do
+      let!(:user) { create(:user, :confirmed, password: password) }
+
+      it 'does get a sucessful answer' do
+        subject
+        expect(response).to be_successful
+      end
+
+      it 'does create a new session' do
+        expect { subject }.to(change { User.last.current_sign_in_at })
```
JuanfraM: `expect { subject }.to change { User.last.current_sign_in_at }`

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  context 'when invalid' do
+    context 'when registered and not confirmed' do
+      let(:password) { 'example' }
+      let!(:user) { create(:user, password: password) }
+
+      it 'does not get a sucessful answer' do
+        subject
+        expect(response).not_to be_successful
+      end
```
JuanfraM, line 37: add an empty line between specs

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  context 'when invalid' do
+    context 'when registered and not confirmed' do
+      let(:password) { 'example' }
+      let!(:user) { create(:user, password: password) }
+
+      it 'does not get a sucessful answer' do
+        subject
+        expect(response).not_to be_successful
```
JuanfraM: This is too vague, we should expect a specific error here

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+      it 'does not sign_in' do
+        subject
+        expect { subject }.not_to(change { User.last.current_sign_in_at })
```
JuanfraM: `expect { subject }.not_to change { User.last.current_sign_in_at }`

spec/request/session/create_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+
+    context 'when registered and confirmed but invalid credentials' do
+      let!(:user) { create(:confirmed_user) }
+      let(:password) { 'another' }
+
+      it 'does not get a sucessful answer' do
+        subject
+        expect(response).not_to be_successful
```
JuanfraM: same as above

spec/request/user/create_spec.rb
```diff
@@ -16,57 +17,152 @@
...
+        it 'does not create a new user' do
+          expect(User.count).to eq(0)
+        end
+
+        it 'returns unprocessable_entity status' do
+          expect(response).to have_http_status(:unprocessable_entity)
+        end
+
+        it 'returns error message' do
+          @data = JSON.parse(response.body)
```
JuanfraM: unnecessary use of `@`

spec/request/user/create_spec.rb
```diff
@@ -16,57 +17,152 @@
...
```
JuanfraM: same as above

app/controllers/api/v1/api_controller.rb
t-romani: I thought it actually was inheriting from ApplicationController which inherited from API, so should I make it 
`class ApiController < ApplicationController`
instead @JuanfraM ?

app/controllers/api/v1/confirmations_controller.rb
t-romani: i dont know... does devise works only with json? the main reason for this was _the include_, which makes sure requests come only in json format. @JuanfraM

app/controllers/api/v1/sessions_controller.rb
t-romani: devise does this automatically, it returns json data which includes id, email and a few more attributes, should I personalize the response or keep devise's one? @JuanfraM

spec/models/user_spec.rb
t-romani: this gives me some issues due to provider being the email itself 😞 need a workaround tip

app/controllers/api/v1/sessions_controller.rb
JuanfraM: the response should be: `user: {full_name: 'bla bla', }`

Gemfile
```diff
@@ -38,12 +38,14 @@ group :development, :test do
...
   gem 'factory_bot_rails', '~> 5.0', '>= 5.0.2'
+  gem 'letter_opener', '~> 1.7'
   gem 'rspec-rails', '~> 3.8', '>= 3.8.2'
 end
 
 group :test do
-  gem 'database_cleaner', '~> 1.7'
+  gem 'faker', '~> 2.1', '>= 2.1.2'
   gem 'rspec-json_expectations'
+  gem 'shoulda-matchers'
```
JuanfraM: version?

JuanfraM (approved):
👌 
Just one minor comment

PR #3, opened 2019-08-19
Feature/sign out

t-romani:


spec/factories/user.rb
```diff
@@ -0,0 +1,14 @@
+FactoryBot.define do
+  factory :user do
+    email     { 'test_user@example.com' }
```
gonzaM: You can use `Faker::Internet.email` here

spec/factories/user.rb
```diff
@@ -0,0 +1,14 @@
...
+    password  { Faker::Internet.password }
+    full_name { Faker::Name.name_with_middle }
+    gender    { 2 }
```
gonzaM: You can use something like: `User.genders.keys.sample` (.sample returns a random element from the array)

spec/models/user_spec.rb
```diff
@@ -1,13 +1,11 @@
 require 'rails_helper'
-RSpec.describe User, :type => :model do
-  subject do
-    @user1 = FactoryBot.create(:user)
-  end
+describe User, :type => :model do
```
gonzaM: Use the new syntax: `type: :model`

spec/models/user_spec.rb
```diff
@@ -1,13 +1,11 @@
...
+  subject { @user1 = create(:user) }
 
   describe '#valid attributes' do
     it 'is valid with valid attributes' do
```
gonzaM: The description should be `'is valid'` then you can use a context '`with valid params`'. So `with valid params... is valid`

spec/request/session/create_spec.rb
```diff
@@ -1,17 +1,18 @@
...
-    post '/api/v1/users/sign_in', params: {
-      email: user_attrs[:user][:email], password: user_attrs[:user][:password]
+    post api_v1_user_session_path, params: {
+      email: user_attrs[:user][:email], password: password
     }, as: :json
   end
 
   context 'when valid' do
+    let(:password) { 'example' }
     context 'when registered and confirmed' do
```
gonzaM, line 13: Add an empty line after the `let`

spec/request/session/create_spec.rb
```diff
@@ -26,7 +27,8 @@
 
   context 'when invalid' do
     context 'when registered and not confirmed' do
-      let!(:user) { FactoryBot.create(:user) }
+      let(:password) { 'example' }
+      let!(:user) { create(:user, password: password) }
```
gonzaM: I would align this

spec/request/session/create_spec.rb
```diff
@@ -39,19 +41,16 @@
     end
 
     context 'when registered and confirmed but invalid credentials' do
-      let!(:user) { FactoryBot.create(:confirmed_user) }
+      let!(:user) { create(:confirmed_user) }
+      let(:password) { 'another' }
```
gonzaM: Same as above (align)

spec/request/session/delete_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'rails_helper'
+
+describe 'POST #delete sign_out', type: :request do
+  let!(:user_attrs) { { user: attributes_for(:user) } }
+  let!(:user) { create(:user, :confirmed) }
+  let!(:auth_headers) { auth_user_headers }
```
gonzaM: I would align this too and also add an empty line after let

spec/request/session/delete_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  subject do
+    delete destroy_api_v1_user_session_path, headers: auth_headers, as: :json
+  end
+
+  context 'when valid' do
+    context 'when signed in' do
```
gonzaM, line 12: There is no context for `when not signed in`?

spec/request/session/delete_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      it 'does get a sucessful answer' do
```
gonzaM: just `gets a ...`? Same for all descriptions

spec/request/session/delete_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        expect { subject }.to change { User.last.tokens.size }.by(-1)
+      end
+    end
+  end
+
+  context 'when invalid' do
+    context 'when not signed in' do
+      it 'does get a not found error' do
+        subject
+        delete destroy_api_v1_user_session_path, headers: auth_headers, as: :json
```
gonzaM: I don't think this is the best way to do this. Use a `before` block if necessary. Or even better, don't log in the user.

spec/request/user/create_spec.rb
```diff
@@ -56,7 +56,7 @@
 
     context 'when invalid password' do
       it '(different) it does not create a new user' do
-        user_params[:user][:password] = 'password'
+        # user_params[:user][:password] = 'password'
```
gonzaM: Remove commented code

gonzaM (approved):
Good job!

spec/request/session/delete_spec.rb
t-romani: there is further on, inside the invalid context

Alxzu (approved):

JuanfraM (approved):
good job!

PR #4, opened 2019-08-20
Feature/edit profile

t-romani:
should be adding raising exception with different status code rather than 500 for when invalid attributes are used for updating

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -1,9 +1,10 @@
 module Api
   module V1
     class ApiController < ApplicationController
-      include DeviseTokenAuth::Concerns::SetUsedByToken
+      include ActAsApi
+      include DeviseTokenAuth::Concerns::SetUserByToken
 
-      before_action :authenticate_user!
+      before_action :authenticate_user!, except: %i[new create]
     end
   end
 end
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -4,7 +4,7 @@ module Api
       include ActAsApi
       include DeviseTokenAuth::Concerns::SetUserByToken
 
-      before_action :authenticate_user!, except: %i[new create]
+      before_action :authenticate_user!, except: %i[create]
 
       rescue_from ArgumentError, with:
         :render_error_response
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -5,6 +5,13 @@ module Api
       include DeviseTokenAuth::Concerns::SetUserByToken
 
       before_action :authenticate_user!, except: %i[new create]
+
+      rescue_from ArgumentError, with:
+        :render_error_response
+
+      def render_error_response(error)
+        render json: { error: error }, status: 400
+      end
     end
   end
 end
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -1,12 +1,15 @@
 module Api
   module V1
     class UsersController < Api::V1::ApiController
+      helper_method :user
+
       def update
-        render_update_success if current_user.update!(user_params)
+        user.update!(user_params)
+        render :show
       end
 
-      def render_update_success
-        render :update
+      def user
+        @user ||= current_user
       end
 
       private
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -1,0 +1,19 @@
+module Api
+  module V1
+    class UsersController < Api::V1::ApiController
+      def update
+        render_update_success if current_user.update!(user_params)
+      end
+
+      def render_update_success
+        render :update
+      end
+
+      private
+
+      def user_params
+        params.require(:user).permit(:full_name, :gender)
+      end
+    end
+  end
+end
--- a/app/views/api/v1/users/_info.json.jbuilder
+++ b/app/views/api/v1/users/_info.json.jbuilder
@@ -1,0 +1,3 @@
+json.id       user.id
+json.name     user.full_name
+json.gender   user.gender
--- a/app/views/api/v1/users/update.json.jbuilder
+++ b/app/views/api/v1/users/update.json.jbuilder
@@ -1,6 +1,0 @@
-json.data do
-  json.user do
-    json.name @current_user.full_name
-    json.gender @current_user.gender
-  end
-end
--- a/app/views/api/v1/users/update.json.jbuilder
+++ b/app/views/api/v1/users/update.json.jbuilder
@@ -1,0 +1,6 @@
+json.data do
+  json.user do
+    json.name @current_user.full_name
+    json.gender @current_user.gender
+  end
+end
--- a/spec/factories/user.rb
+++ b/spec/factories/user.rb
@@ -4,9 +4,6 @@ FactoryBot.define do
     password  { Faker::Internet.password }
     full_name { Faker::Name.name_with_middle }
     gender    { User.genders.keys.sample }
-
-    trait :confirmed do
-      after :create, &:confirm
-    end
+    after :create, &:confirm
   end
 end
--- a/spec/request/user/update_spec.rb
+++ b/spec/request/user/update_spec.rb
@@ -1,0 +1,69 @@
+require 'rails_helper'
+
+describe 'PATCH #update user params', type: :request do
+  let!(:user) { create(:user, :confirmed, gender: 'male') }
+  let!(:name) { Faker::Name.name_with_middle }
+  let!(:last_user) { User.last }
+  subject do
+    patch api_v1_user_path(id: user.id), params: user_params,
+                                         headers: auth_headers,
+                                         as: :json
+  end
+
+  context 'when valid' do
+    let!(:auth_headers) { auth_user_headers }
+    context 'when changing name' do
+      let(:user_params) { { user: { full_name: name } } }
+
+      it 'does get a successful answer' do
+        subject
+        expect(response).to be_successful
+      end
+
+      it 'does change the name' do
+        expect { subject }.to(change { User.last.full_name })
+      end
+    end
+    context 'when changing gender' do
+      let(:user_params) { { user: { gender: 'female' } } }
+      it 'does get a successful answer' do
+        subject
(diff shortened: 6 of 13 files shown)


t-romani:
Probably gonna work a bit more on catching that exception

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Api
+  module V1
+    class UsersController < Api::V1::ApiController
+      def update
+        current_user.update!(user_params)
```
Alxzu: Usually, a good approach when updating a record it's returning `head :no_content`. You can [take a look at this](https://stackoverflow.com/questions/14716151/why-does-rails-want-to-return-head-no-content-for-json-put-requests) where this behavior is discussed briefly

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      end
+
+      def render_update_success
+        render json: { user: current_user }
```
Alxzu: This is ok, but I think it would be better to have a view (JBuilder view) to render this kind of stuff, it's better since you can:
1. Have one place to render users
2. Customize response

app/controllers/api/v1/users_controller.rb
t-romani: Excellent! To make sure, can I consider PUT and PATCH as the same or similar kind of request? In this terms at least @Alxzu

app/controllers/api/v1/users_controller.rb
t-romani: Great, @JuanfraM talked to me about this, gonna check it out! Tnx!

app/controllers/api/v1/users_controller.rb
Alxzu: Yes, actually in REST you can handle edit requests with both 👍

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Api
+  module V1
+    class UsersController < Api::V1::ApiController
+      def update
+        render_update_success if current_user.update!(user_params)
+      end
+
+      def render_update_success
```
gonzaM: I think this is not necessary. Just:
```ruby
def update
  current_user.update!(user_params)
end
```
will render the update if no errors.

app/views/api/v1/users/update.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+json.data do
```
gonzaM: Is the frontend expecting this key? If not, I would remove it.

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'rails_helper'
+
+describe 'PATCH #update user params', type: :request do
+  let!(:user) { create(:user, :confirmed) }
```
gonzaM: I think you should add the confirmed as default, so you don't need to add the trait every time you use a user.
Generally, the user that you need, should be confirmed, right?
You can create a `trait` for not confirmed ones. WDYT?

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  let!(:name) { Faker::Name.name_with_middle }
+  let!(:last_user) { User.last }
```
gonzaM: I think you don't need this

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  subject do
```
gonzaM, line 8: I would align the lets and also add an empty line after lets

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    patch api_v1_user_path(id: user.id), params: user_params,
+                                         headers: auth_headers,
+                                         as: :json
+  end
+
+  context 'when valid' do
+    let!(:auth_headers) { auth_user_headers }
+    context 'when changing name' do
```
gonzaM, line 15: Add empty line after let

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      let(:user_params) { { user: { full_name: name } } }
+
+      it 'does get a successful answer' do
+        subject
+        expect(response).to be_successful
+      end
+
+      it 'does change the name' do
+        expect { subject }.to(change { User.last.full_name })
```
gonzaM: Use `user` instead of `User.last`

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+    context 'when changing gender' do
+      let(:user_params) { { user: { gender: 'female' } } }
+      it 'does get a successful answer' do
+        subject
+        expect(response).to be_successful
+      end
+
+      it 'does change the gender' do
+        expect { subject }.to(change { User.last.gender })
```
gonzaM: Same as above

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+    end
+  end
+
+  context 'when invalid' do
+    let(:user_params) { { user: { full_name: name, gender: 'female' } } }
+    context 'when signed out' do
```
gonzaM, line 49: Add empty line after let

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        expect(response).to have_http_status(:unauthorized)
+      end
+
+      it 'does not change the name/gender' do
+        expect { subject }.not_to(change { User.last.attributes})
+      end
+    end
+
+    context 'when signed in' do
+      let!(:auth_headers) { auth_user_headers }
```
gonzaM: I think you can move this above to avoid repeating it. You can redefine it only in the "signed out" context.
Also, I think it's not necessary to test for all requests what happen if the user is not logged in.

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      context 'when gender invalid' do
```
gonzaM, line 63: empty line

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        let(:user_params) { { user: { gender: 'invalid' } } }
+        it 'does get a bad request answer' do
```
gonzaM, line 66: empty line

spec/request/user/update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+          subject
+          expect(response).to have_http_status(:bad_request)
+        end
+
+        it 'does not change the name/gender' do
+          expect { subject }.not_to(change { User.last.attributes })
```
gonzaM: I would check the specific attributes

spec/request/user/update_spec.rb
t-romani: Great, how about, insted of a `trait`, a before block that works for that context?
The trait might be... tricky. @gonzaM

gonzaM: Do you mean a context with a user with the confirmed_at nil? 👍

spec/request/user/update_spec.rb
t-romani: @gonzaM  what do you mean by 

> it's not necessary to test for all requests what happen if the user is not logged in

from what I understand, there are only 2 tests regarding what happens if the user is not logged in, being response status and no attributes changed (both at the same time).

gonzaM (approved):
Good job!

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Api
+  module V1
+    class UsersController < Api::V1::ApiController
+      def update
+        @user = current_user
```
jiv: You can memoize this for future methods and to keep the `update` cleaner

jiv (approved):

app/controllers/api/v1/api_controller.rb
```diff
@@ -1,9 +1,17 @@
 module Api
   module V1
     class ApiController < ApplicationController
-      include DeviseTokenAuth::Concerns::SetUsedByToken
+      include ActAsApi
+      include DeviseTokenAuth::Concerns::SetUserByToken
 
-      before_action :authenticate_user!
+      before_action :authenticate_user!, except: %i[new create]
```
JuanfraM: are you using `new` action?

app/controllers/api/v1/users_controller.rb
JuanfraM: I would make this change as @jiv suggests

[Memoization](https://thoughtbot.com/blog/ruby-memoization-and-alternatives)

app/views/api/v1/users/_user.json.jbuilder
```diff
@@ -0,0 +1,4 @@
+json.user do
+  json.name @user.full_name
+  json.gender @user.gender
+end
```
JuanfraM: Just to keep the same convention we usually name this file as `_info.json.jbuilder` since you are into the folder `users`

spec/factories/user.rb
```diff
@@ -4,9 +4,10 @@
     password  { Faker::Internet.password }
     full_name { Faker::Name.name_with_middle }
     gender    { User.genders.keys.sample }
+    after :create, &:confirm
 
-    trait :confirmed do
-      after :create, &:confirm
+    trait :not_confirmed do
```
JuanfraM: I think you are not using this anymore

JuanfraM (approved):
👌 
Approve after comments are fixed

app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Api
+  module V1
+    class UsersController < Api::V1::ApiController
+      def update
+        @user ||= current_user
+        @user.update!(user_params)
+        render :show
+      end
```
JuanfraM, line 9: helper method with memoization called `user`?

app/views/api/v1/users/_info.json.jbuilder
```diff
@@ -0,0 +1,4 @@
+json.user do
```
JuanfraM: I prefer you just return the data here and use the variable user in the show view

PR #5, opened 2019-08-27
Feature/topics

t-romani:


app/controllers/api/v1/topics_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Api
+  module V1
+    class TopicsController < Api::V1::ApiController
+      def index
+        @topics = Topic.all
```
gonzaM, line 5: Should you use pagination here?

app/models/topic.rb
```diff
@@ -0,0 +1,29 @@
...
+#
+#  index_topics_on_title  (title) UNIQUE
+#
+
+class Topic < ApplicationRecord
+  include ActiveStorageSupport::SupportForBase64
+
+  has_one_base64_attached :image
+
+  validates :title, presence: true, uniqueness: true
```
gonzaM: Case sensitive uniqueness? If not, you can add `case_sensitive: false`

app/models/topic.rb
```diff
@@ -0,0 +1,29 @@
...
+  validate :image_presence
+
+  def image_presence
+    errors.add(:attached_image, 'no image added') unless image.attached?
+  end
+
+  # include ActiveStorageValidations::AttachedValidator
+  # validates :image, attached: true
```
gonzaM: Remove commented code

app/views/api/v1/topics/_info.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+json.id          topic.id
+json.title       topic.title
+json.image       topic.image
+json.created_at  topic.created_at
+json.updated_at  topic.updated_at
```
gonzaM: Maybe you can use `extract!`? https://www.rubydoc.info/github/rails/jbuilder/Jbuilder:extract!

db/seeds.rb
```diff
@@ -5,3 +5,26 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+t = Topic.create(
+  'id': 7,
+  'title': 'The Pragmatic Programmer',
+  'created_at': '2019-08-26T15:27:50.113Z',
```
gonzaM: The id, created_at, and updated_at are not necessary. You can do something like:
```ruby
titles = ['The Pragmatic Programmer', 'The Stupendous Origin', ...]
titles.each do |title|
  t = Topic.create(title: title)
  t.image.attach...
end
```

spec/factories/topic.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :topic do
+    title     { Faker::Verb.base }
+    image     {
```
gonzaM: Use `do ... end` when is a multiline block

spec/models/topic_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+#  index_topics_on_title  (title) UNIQUE
+#
+
+require 'rails_helper'
+describe Topic, type: :model do
+  subject { build(:topic) }
+
+  describe 'presence validations' do
+    it { should validate_presence_of(:title) }
```
gonzaM: Use the new syntax: `is_expected.to validate_presence_of(:title)`

spec/models/topic_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+# Indexes
+#
+#  index_topics_on_title  (title) UNIQUE
+#
+
+require 'rails_helper'
+describe Topic, type: :model do
+  subject { build(:topic) }
+
+  describe 'presence validations' do
```
gonzaM: I think it's better to have a describe `'validations'` and check all validation inside (and remove this one)

spec/models/topic_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      let!(:topic) { build(:topic, image: nil) }
+
+      it 'isnt valid' do
+        expect(topic).not_to be_valid
+      end
+    end
+  end
+
+  describe 'uniqueness' do
+    it { should validate_uniqueness_of(:title) }
```
gonzaM: Same here. Use the new syntax and I would move to a 'validations' block

spec/models/user_spec.rb
```diff
@@ -1,3 +1,37 @@
...
+#  created_at             :datetime         not null
+#  updated_at             :datetime         not null
+#
+# Indexes
+#
+#  index_users_on_email                 (email) UNIQUE
+#  index_users_on_reset_password_token  (reset_password_token) UNIQUE
+#  index_users_on_uid_and_provider      (uid,provider) UNIQUE
+#
+
```
gonzaM, line 34: Check here. In this class you use a `describe 'validations'`.
Also, you should use the new syntax in all the code. There is a [gem](https://github.com/yujinakayama/transpec) that can help you with that if you want.

spec/request/topic/index_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  let!(:topic2) { create(:topic) }
+  let(:topics)  { Topic.all }
+
+  subject do
+    get api_v1_topics_path, headers: auth_headers, as: :json
+  end
+
+  before do
+    subject
+  end
```
gonzaM: You can do `before { subject }`

spec/request/topic/index_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    context 'when logged in' do
+      let!(:user)         { create(:user) }
+      let!(:auth_headers) { auth_user_headers }
+
+      it 'gets a successfull answer' do
+        expect(response).to be_successful
+      end
+
+      it 'returns list of topics' do
+        data = JSON.parse(response.body)
```
gonzaM: I think you can move this to a helper so you can use it in all responses.

spec/request/topic/index_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  subject do
+    get api_v1_topics_path, headers: auth_headers, as: :json
+  end
+
+  before do
+    subject
+  end
+
+  context 'when valid' do
+    context 'when logged in' do
```
gonzaM: I don't think this context is necessary here

gonzaM (commented):
Good job! 👏

app/models/topic.rb
t-romani: waiting for @JuanfraM to have a talk about that gem 👀

app/views/api/v1/topics/_info.json.jbuilder
t-romani: @JuanfraM showed me this way.
Which one is the office standard? @gonzaM @JuanfraM

gonzaM: I don't think there is a standard 😄 
The `extract!` is easy to use and also you don't need to repeat code since in this case, all the JSON keys are the same as the attributes' names.
When you need to add custom JSON keys, you can use the other approach. 
I wanted to show you this alternative but is up to you. Both are okay.

JuanfraM: @t-romani had done it as you mentioned 😅 . I told her I've seen this way many times but I think it's ok both. Also, if the JSON keys are the same as attributes what @gonzaM said is more appropriate

app/controllers/api/v1/topics_controller.rb
t-romani: @gonzaM  yes, but pagination is part of a card I will do further on

Gemfile
```diff
@@ -3,6 +3,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby '2.6.3'
 
+gem 'active_storage_validations', '~> 0.8.2'
```
JuanfraM: are you going to use this?

app/models/topic.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+
+class Topic < ApplicationRecord
+  include ActiveStorageSupport::SupportForBase64
+
+  has_one_base64_attached :image
+
+  validates :title, presence: true, uniqueness: { case_sensitive: false }
+
+  after_commit :add_default_image, on: %i[create]
```
JuanfraM: why not before save?

db/migrate/20190826140802_add_index_to_topics.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToTopics < ActiveRecord::Migration[5.2]
+  def change
+    add_index :topics, %i[title], unique: true
```
JuanfraM: Array is unnecessary in this case, just: `add_index :topics, :title, unique: true`

JuanfraM: Also, couldn't you set the index in the migration above?

db/seeds.rb
```diff
@@ -5,3 +5,7 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+titles = ['The Pragmatic Programmer', 'The Stupendous Origin', 'The Awakening of the Sencha']
+titles.each do |title|
+  t = Topic.create(title: title)
```
JuanfraM: if you use `find_or_create_by ` instead of just `create` you can run the seed again in the same database in case you want to add new data

spec/helpers.rb
```diff
@@ -1,3 +1,7 @@
 def auth_user_headers
   user.create_new_auth_token
 end
+
+def parsed_data
+  data = JSON.parse(response.body)
+end
```
JuanfraM: nice!

spec/models/user_spec.rb
JuanfraM: Also, you can always check [BetterSpecs](http://www.betterspecs.org/#expect) as reference

spec/request/topic/index_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'rails_helper'
+
+describe 'GET #index topic', type: :request do
+  let!(:topic1) { create(:topic) }
+  let!(:topic2) { create(:topic) }
+  let(:topics)  { Topic.all }
```
JuanfraM: these 3 lines above can be wroten as:
`let(:topics)  { create_list(:topic, 2) }`

spec/support/factory_bot.rb
```diff
@@ -1,4 +1,6 @@
 #Factory Bot configuration
   RSpec.configure do |config|
+    include ActionDispatch::TestProcess
```
JuanfraM: what is this used for?

app/models/topic.rb
t-romani: Probably because the guy from stackoverflow wrote it that way, and surely that was the one and only way to do it (lol no)

db/migrate/20190826140802_add_index_to_topics.rb
t-romani: I definitely can do that, I did have some issues tho, in the next PR when trying to set foreign key from topics on targets in the same migration as the creation of the targets table. Also I'm not entirely sure that must be done. We should talk about it @JuanfraM

spec/support/factory_bot.rb
t-romani: @JuanfraM it enabled me able to upload an image on the factory.
Since now I got the default image, I don't see the use anymore, cause image uploads can be made through the model.

app/views/api/v1/topics/_info.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+json.id          topic.id
+json.title       topic.title
+json.image       topic.image
+json.created_at  topic.created_at
```
JuanfraM: I think timestamps are not necessary

JuanfraM (approved):
Just one minor comment
Nice job with images!

spec/helpers.rb
gonzaM: missing eof

gonzaM (approved):

PR #7, opened 2019-08-30
target w/index

t-romani:


app/controllers/api/v1/targets_controller.rb
```diff
@@ -9,6 +9,10 @@ def create
         render :show
       end
 
+      def index
+        @targets = user.targets.index
```
gonzaM: What is `.index` for?

gonzaM (commented):
Good job!

app/controllers/api/v1/targets_controller.rb
t-romani: I guess it is not needed then 😆

spec/request/target/index_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+describe 'GET #index target', type: :request do  
+  let!(:user)     { create(:user) }
+  let!(:targets)  { create_list(:target, 2, user: user) }
+  let(:data)      { parsed_data }
```
gonzaM: where are you using this? 🤔

gonzaM (approved):

spec/request/target/index_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  context 'when valid' do
+    let!(:auth_headers) { auth_user_headers }
+
+    it 'gets a successfull answer' do
+      expect(response).to be_successful
+    end
+
+    it 'returns list of targets' do
+      expect(parsed_data['targets'].count).to eq(targets.count)
+    end
```
JuanfraM, line 23: The title of the spec is incorrect here, you are testing the correct amount of target.
I'd also test that the `ids` are the same as expected using `match_array `

JuanfraM (approved):

PR #8, opened 2019-08-30
targets w/delete action

t-romani:


app/controllers/api/v1/targets_controller.rb
```diff
@@ -12,6 +12,11 @@ def index
         @targets = user.targets
       end
 
+      def destroy
+        @target = user.targets.find(params[:id])
+        @target.destroy
```
gonzaM: I would use `destroy!`, also I don't think it's necessary to use the `@target`. So `user.targets.find(params[:id]).destroy!` should be ok

config/locales/en.yml
```diff
@@ -35,6 +35,7 @@ en:
     error: 
       invalid_request:
         content_type: 'Invalid Content Type'
+        content_not_found: "Couldn't find that id"
```
gonzaM: I would use "Couldn't find the record" so is more general

spec/request/target/delete_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  let!(:target)         { create(:target, user: user) }
+  let(:data)            { parsed_data }
+  let(:target_id)       { target.id }
+
+  subject do
+    delete api_v1_target_path(id: target_id), headers: auth_headers, as: :json
+  end
+
+  context 'when valid' do
+    context 'when signed in' do
```
gonzaM, line 14: Seems that you always add this context, but I don't think it's necessary.

spec/request/target/delete_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      it 'does not modify targets' do
+        expect { subject }.not_to change(user.targets.reload, :count)
+      end
+    end
+
+    context 'when invalid target id' do
+      let!(:auth_headers)    { auth_user_headers }
+      let!(:another_user)    { create(:user) }
```
gonzaM: I don't think this is necessary. In the factory below, don't specify the user, and the factory will create another user.

gonzaM (commented):
Small comments.
Good job!

gonzaM (approved):

app/controllers/api/v1/api_controller.rb
```diff
@@ -10,10 +10,17 @@ class ApiController < ApplicationController
...
+      rescue_from ActiveRecord::RecordNotFound, with:
+        :render_not_found_response
 
       def render_error_response(error)
         render json: { error: error }, status: 400
       end
+
+      def render_not_found_response(error)
+        render json: { error: I18n.t('api.error.invalid_request.content_not_found')
+          }, status: 404
```
JuanfraM: `status: :not_found`

[Rails HTTP Status Codes](http://www.railsstatuscodes.com/l)

Also you can edit `400` above

app/controllers/api/v1/targets_controller.rb
```diff
@@ -12,6 +12,10 @@ def index
         @targets = user.targets
       end
 
+      def destroy
+        user.targets.find(params[:id]).destroy!
+      end
```
JuanfraM, line 17: Since this controller is about targets and you are using `user.targets` three times, you can create a method called `targets` with memoization as you did with user,  and reutilize it on destroy, create and index. Also if you set this method as a helper method then the action item would be just `def index; end`

spec/request/target/delete_spec.rb
JuanfraM: As @gonzaM said, I think it's ok just with the first context

spec/request/target/delete_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        expect { subject }.to change(user.targets.reload, :count).by(-1)
+      end
+    end
+  end
+
+  context 'when invalid' do
+    let!(:auth_headers) { {} }
+    it 'gets a unauthorized error' do
+      subject
+      expect(response).to have_http_status(401)
```
JuanfraM: rails code here

spec/request/target/delete_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  context 'when invalid target id' do
+    let!(:auth_headers)    { auth_user_headers }
+    let!(:another_target)  { create(:target) }
+    let!(:target_id)       { { id: another_target.id } }
+
+    it 'gets not_found status' do
+      subject
+      expect(response).to have_http_status(404)
```
JuanfraM: same as above

app/controllers/api/v1/targets_controller.rb
t-romani: I don't think defining index like that works. Spec failed when returning the list of targets.
If I call the memo inside the function it works. @JuanfraM

jiv (approved):

JuanfraM (approved):

PR #9, opened 2019-09-02
Feature/targets maximum

t-romani:
the limit of targets per user is 10
after creating 10 targets the user should't be able to create another target, and receive an error.

### DIFF
--- a/app/controllers/api/v1/targets_controller.rb
+++ b/app/controllers/api/v1/targets_controller.rb
@@ -20,6 +20,11 @@ module Api
         user.targets.find(params[:id]).destroy!
       end
 
+      def destroy
+        @target = user.targets.find(params[:id])
+        @target.destroy
+      end
+
       def user
         @user ||= current_user
       end
--- a/app/controllers/api/v1/targets_controller.rb
+++ b/app/controllers/api/v1/targets_controller.rb
@@ -20,11 +20,6 @@ module Api
         user.targets.find(params[:id]).destroy!
       end
 
-      def destroy
-        @target = user.targets.find(params[:id])
-        @target.destroy
-      end
-
       def user
         @user ||= current_user
       end
--- a/app/controllers/api/v1/targets_controller.rb
+++ b/app/controllers/api/v1/targets_controller.rb
@@ -16,10 +16,6 @@ module Api
         targets.find(params[:id]).destroy!
       end
 
-      def destroy
-        user.targets.find(params[:id]).destroy!
-      end
-
       def user
         @user ||= current_user
       end
--- a/app/controllers/api/v1/targets_controller.rb
+++ b/app/controllers/api/v1/targets_controller.rb
@@ -16,6 +16,10 @@ module Api
         targets.find(params[:id]).destroy!
       end
 
+      def destroy
+        user.targets.find(params[:id]).destroy!
+      end
+
       def user
         @user ||= current_user
       end
--- a/app/errors/TargetLimitReached.rb
+++ b/app/errors/TargetLimitReached.rb
@@ -1,1 +1,0 @@
-class TargetLimitReached < StandardError; end
--- a/app/errors/TargetLimitReached.rb
+++ b/app/errors/TargetLimitReached.rb
@@ -1,0 +1,1 @@
+class TargetLimitReached < StandardError; end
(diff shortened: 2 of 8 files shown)


app/models/target.rb
```diff
@@ -32,4 +32,12 @@ class Target < ApplicationRecord
   validates :title, presence: true, uniqueness: {
     scope: %i[topic_id user_id], case_sensitive: false
   }
+
+  before_save :limit_number_of_targets
```
gonzaM: Can't you use a validation?

t-romani: Couldn't come up with anything, any ideas or useful links @gonzaM ?

gonzaM: Do you know custom validations? https://guides.rubyonrails.org/active_record_validations.html#custom-methods
Something like:
```ruby
validate :limit_number_of_targets

def limit_number_of_targets
  return if user has less than 10
  errors.add(:base, "A custom error...")
end
```

app/errors/TargetLimitReached.rb
```diff
@@ -0,0 +1 @@
+class TargetLimitReached < StandardError; end
```
JuanfraM: if you call this file `target_limit_reached_exception.rb` then you need to import it in the api controller?

spec/request/target/create_spec.rb
```diff
@@ -71,5 +70,24 @@
         end
       end
     end
+
+    context 'when limit reached' do
+      before do
+        10.times do
+          user.targets.create!(attributes_for(:target, topic_id: topic.id))
+        end
```
JuanfraM: use `create_list` for this purpose from [FactoryBot](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md)

app/models/target.rb
JuanfraM: I think what @gonzaM said is more appropriate in this case. This validation only apply in the target model

JuanfraM (commented):
Nice job!
I left some comments

spec/request/target/create_spec.rb
```diff
@@ -71,5 +70,24 @@
...
+      end
```
andresg4, line 78: Maybe here you can use the `create_list` from `FactoryBot` that lets you create multiple records ([create_list](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#building-or-creating-multiple-records)).
You could do something like this:
```ruby
before do
  create_list(:target, 10, user: user, topic: topic)
  subject
end
```
And with this you can also avoid the `subject` in each test.

app/errors/TargetLimitReached.rb
t-romani: this is no longer in the proyect

app/models/target.rb
t-romani: this is creating me some issues with the creation of targets when the user is null, for example on belong_to spec validation.

app/models/target.rb
```diff
@@ -32,4 +32,15 @@ class Target < ApplicationRecord
   validates :title, presence: true, uniqueness: {
     scope: %i[topic_id user_id], case_sensitive: false
   }
+
+  validate :limit_number_of_targets
+
+  private
+
+  def limit_number_of_targets
+    return if user && user.targets.count < 10
```
gonzaM, line 41: <=?

gonzaM (approved):

spec/request/target/create_spec.rb
```diff
@@ -71,5 +70,21 @@
...
       end
     end
+
+    context 'when limit reached' do
+      before do
+        create_list(:target, 10, user: user, topic: topic)
+        subject
+      end
+
+      it 'gets an unauthorized response' do
```
andresg4: Maybe `'gets a bad request response'`?

andresg4 (approved):
👏

app/models/target.rb
t-romani: I think < works, because the limit is exactly 10

spec/request/target/create_spec.rb
t-romani: 🤦‍♀

JuanfraM (approved):

app/controllers/api/v1/targets_controller.rb
```diff
@@ -16,6 +16,10 @@ def destroy
         targets.find(params[:id]).destroy!
       end
 
+      def destroy
+        user.targets.find(params[:id]).destroy!
```
JuanfraM: You've already declared the helper `targets`, you can use it here.
I think it was added in the PR of delete targets but it seems it was overridden here.

JuanfraM (approved):
Just one small comment

PR #11, opened 2019-09-06
Feature/notifications

t-romani:
big pr;
when user logs in for the first time, meaning has confirmed his email, a player_id, needed by onesignal is assigned, so later email notifications can be send.
when user creates a target that matches in radius with other targets, both users receive email notification through onesignal, this is done in a job, since it has not high priority


app/jobs/notify_request_job.rb
```diff
@@ -0,0 +1,8 @@
+class NotifyRequestJob < ApplicationJob
+  queue_as :default
+
+  def perform(matching_users)
+    NotificationService.new(matching_users)
+    NotificationService.notify
```
JuanfraM: is this service taking the same you've created above?

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -8,8 +8,15 @@ def sign_in_params
       end
 
       def render_create_success
+        set_player_id
         render json: { user: resource_data }
       end
+
+      private
+
+      def set_player_id
```
JuanfraM, line 17: maybe this method could be called register in notification service or something like that so you can reuse it in case other service use something different than a player id

app/models/target.rb
```diff
@@ -30,12 +30,34 @@ class Target < ApplicationRecord
...
+
+  def check_matching_targets
+    matching_users = []
+    Target.find_each do |each_target|
+      matching_users.push(each_target.user) if
+      each_target.topic_id == topic_id &&
+      each_target.user_id != user_id &&
+      matching_users.exclude?(each_target.user) &&
+      distance_to(each_target) <= radius + each_target.radius
+    end
```
JuanfraM, line 56: you can use `map` here and return each element you are looking for. 
Also, you can improve the search of targets using `where` and specifying the ones which have user_id different as the current target

app/models/target.rb
```diff
@@ -30,12 +30,34 @@ class Target < ApplicationRecord
...
+
+    matching_users.push(user) unless matching_users.empty?
+    notificate_users(matching_users) unless matching_users.empty?
```
JuanfraM: you can use `return if matching_users.empty?` and keep the methods below the sentence

app/services/notification_service.rb
```diff
@@ -0,0 +1,25 @@
+class NotificationService
+  def initialize(matching_users)
+    @@users = matching_users
```
JuanfraM: why are you using a class variable here?
you can use [attr_reader](https://www.rubyguides.com/2018/11/attr_accessor/)

app/services/notification_service.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def self.notify
+    ids = []
+    @@users.collect(&:player_id).each do |pid|
```
JuanfraM: collect it's the same as map, I think that map is most used. Just to let you know.
Also, use player_id instead of pid, it's clearer

app/services/notification_service.rb
```diff
@@ -0,0 +1,25 @@
...
+      ids << pid
+    end
+    params = { 'app_id' => ENV['APP_ID'],
+               'email_subject' => 'New Match!',
+               'email_body' => 'You have matched another Target!',
```
JuanfraM: these variables should be in a locale file

app/services/notification_service.rb
```diff
@@ -0,0 +1,25 @@
...
+               'include_player_ids' => ids }
+    uri = URI.parse('https://onesignal.com/api/v1/notifications')
```
JuanfraM: this url should be in an environment constant

app/services/notification_service.rb
```diff
@@ -0,0 +1,25 @@
...
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
```
JuanfraM, line 67: maybe you can create a private method called `send_notification` and move the logic of the request

app/services/player_ids_service.rb
```diff
@@ -0,0 +1,29 @@
+class PlayerIdsService
+  def initialize(user)
+    @user = user
+  end
+
+  def set_id()
+    return unless @user.player_id.nil?
```
JuanfraM: you can use `attr_reader` as I mentioned above

app/services/player_ids_service.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    email_auth_hash = OpenSSL::HMAC.hexdigest('sha256', ENV['API_KEY'], @user.email)
+    email_create_params = {
+      device_type: 11,
+      app_id: ENV['APP_ID'],
+      identifier: @user.email,
+      email_auth_hash: email_auth_hash
+    }
+    uri = URI.parse('https://onesignal.com/api/v1/players')
```
JuanfraM: same as the comments above with the variable and the method

config/application.yml.example
```diff
@@ -8,4 +8,8 @@ DEVELOP_DATABASE:
 TEST_DATABASE:
 PRODUCTION_DATABASE:
 DBUSERNAME:
-DBPASSWORD:
\ No newline at end of file
+DBPASSWORD:
+
+USER_AUTH_KEY:
+API_KEY:
```
JuanfraM: api_key of what 🤔 , same as app_id. If these variable are for one signal you should use something like:
ONE_SIGNAL_API_KEY , same for the app id

spec/request/session/create_spec.rb
```diff
@@ -9,8 +9,15 @@
 
   context 'when valid' do
     let(:password) { 'example' }
-    context 'when registered and confirmed' do
-      let!(:user) { create(:user, password: password) }
+
+    before do
+      WebMock.stub_request(:post, 'https://onesignal.com/api/v1/players')
+             .to_return(status: 200,
+                        body: File.new('spec/support/fixtures/new_player_id.json'))
```
JuanfraM: maybe this file should be called `new_player_id_success.json` because it's the success case

spec/request/target/create_spec.rb
```diff
@@ -31,6 +31,28 @@
...
+               topic_id: topic.id,
+               user: another_user,
+               latitude: target_params[:target][:latitude],
+               longitude: target_params[:target][:longitude])
+      end
+
+      before do
+        WebMock.stub_request(:post, 'https://onesignal.com/api/v1/notifications')
+               .to_return(status: 200,
+                          body: File.new('spec/support/fixtures/new_target_match.json'))
```
JuanfraM: same as above

app/jobs/notify_request_job.rb
t-romani: yes, it becomes a class variable. Is it wrong?

app/models/target.rb
JuanfraM: since map returns an array you can use it and declare:
`matching_users = Target.where(user_id: user.id)........`
then you don't need to declare `matching_users = []`

app/models/target.rb
```diff
@@ -30,12 +30,42 @@ class Target < ApplicationRecord
...
+
+  def notificate_users(matching_users)
+    NotifyRequestJob.perform_later(matching_users)
+  end
+
+  def matchs(matching_users, target)
+    return true if
+    target.topic_id == topic_id &&
+    matching_users.exclude?(target.user) &&
+    distance_to(target) <= radius + target.radius
```
JuanfraM: can you just return the sentences? I think it has already the value you are looking for

app/services/notification_service.rb
```diff
@@ -0,0 +1,32 @@
+class NotificationService
+  attr_accessor :users
+
+  def initialize(matching_users)
+    @users = matching_users
+  end
+
+  def notify
+    ids = @users.map(&:player_id)
```
JuanfraM: I think you don't  need the `@` here because you define it with `attr_accessor`

app/services/notification_service.rb
```diff
@@ -0,0 +1,32 @@
...
+    params = request_params(ids)
+    uri = URI.parse(ENV['ONE_SIGNAL_NOTIFY_URL'].to_s)
+    send_request(uri, params)
+  end
+
+  def request_params(ids)
+    { 'app_id' => ENV['ONE_SIGNAL_APP_ID'],
```
JuanfraM: use the new syntax for hashes.
Also you can align like:
```
{
  name: 'tiziana',
  age: 24
}
```
Also I think the key does not need the `' '`

app/services/notification_service.rb
```diff
@@ -0,0 +1,32 @@
...
+      'email_subject' => I18n.t('api.email.matched_target.subject'),
+      'email_body' => I18n.t('api.email.matched_target.body'),
+      'include_player_ids' => ids }
+  end
+
+  def send_request(uri, params)
```
JuanfraM, line 65: private?

app/services/player_ids_service.rb
```diff
@@ -0,0 +1,48 @@
+class PlayerIdsService
+  attr_accessor :user
+
+  def initialize(user)
+    @user = user
+  end
+
+  def register_for_notification
+    return unless @user.player_id.nil?
```
JuanfraM: same as above

app/services/player_ids_service.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    uri = URI.parse(ENV['ONE_SIGNAL_ID_URL'].to_s)
+    send_request(uri)
+  end
+
+  def send_request(uri)
```
JuanfraM: private?

config/application.yml.example
JuanfraM: ONE_SIGNAL_API_KEY  same for user_auth_key

app/models/target.rb
t-romani: but i need the extra condition checked on the if inside the map, 
if i dont declare what matching users is, it doesnt allow me to call the function match throwing undefined variable `matching_users`.

app/services/notification_service.rb
t-romani: the key definitely needs `' '`

app/models/target.rb
```diff
@@ -30,12 +30,39 @@ class Target < ApplicationRecord
...
 
-    errors.add(:base, I18n
-      .t('api.error.invalid_request.max_target_limit'))
+    errors.add(:base,
+               I18n.t('api.error.invalid_request.max_target_limit'))
+  end
+
+  def check_matching_targets
+    matching_users = []
+    Target.where("user_id != #{user.id}").map do |each_target|
```
jiv: I think `|target|` is fine here, no need for `each` prefix

app/models/target.rb
```diff
@@ -30,12 +30,39 @@ class Target < ApplicationRecord
...
+
+    matching_users.push(user)
+    notificate_users(matching_users)
+  end
+
+  def notificate_users(matching_users)
+    NotifyRequestJob.perform_later(matching_users)
+  end
+
+  def matchs?(matching_users, target)
```
jiv: `matches`

app/services/notification_service.rb
```diff
@@ -0,0 +1,34 @@
+class NotificationService
+  attr_accessor :users
+
+  def initialize(matching_users)
+    @users = matching_users
+  end
+
+  def notify
+    ids = users.map(&:player_id)
```
jiv: I would name this variable `player_ids` instead

app/services/notification_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    params = request_params(ids)
+    uri = URI.parse(ENV['ONE_SIGNAL_NOTIFY_URL'].to_s)
+    send_request(uri, params)
+  end
+
+  private
+
+  def request_params(ids)
```
jiv: same here: `request_params(player_ids)`

app/services/player_ids_service.rb
```diff
@@ -0,0 +1,50 @@
+class PlayerIdsService
```
jiv: I agree with @JuanfraM , all this code could be part of the NotificationService. In the initialize you can pass the specific user too and move the register method in there. If you do this, you need to add an if clause in the `notify` method to return when the matching_users list is empty.

spec/spec_helper.rb
```diff
@@ -15,6 +15,8 @@
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 require 'rspec/json_expectations'
 
+
```
jiv: no need for these new lines

app/models/target.rb
```diff
@@ -30,12 +30,39 @@ class Target < ApplicationRecord
...
 
-    errors.add(:base, I18n
-      .t('api.error.invalid_request.max_target_limit'))
+    errors.add(:base,
+               I18n.t('api.error.invalid_request.max_target_limit'))
+  end
+
+  def check_matching_targets
+    matching_users = []
+    Target.where("user_id != #{user.id}").map do |target|
```
gonzaM: You can use something like: `Target.where.not(user_id: user.id)`. Also, I would move this to a scope.
Also, you can add in the where the topic for the target so you don't need to load all targets: `Target.where.not(user_id: user.id).where(topic_id: topic_id)`

app/models/target.rb
```diff
@@ -30,12 +30,39 @@ class Target < ApplicationRecord
...
+  end
+
+  def notificate_users(matching_users)
+    NotifyRequestJob.perform_later(matching_users)
+  end
+
+  def matches?(matching_users, target)
+    target.topic_id == topic_id &&
+      matching_users.exclude?(target.user) &&
+      distance_to(target) <= radius + target.radius
```
gonzaM, line 70: Can't you use a method like near or another one, so you avoid making one query for each one? Maybe it's not possible, but could you just in case?

jiv (approved):

app/services/notification_service.rb
```diff
@@ -0,0 +1,76 @@
+class NotificationService
+  attr_accessor :user, :matching_users
+
+  def initialize(user, matching_users)
+    @user = user
+    @matching_users = matching_users
+
+    set_id unless user.nil?
```
gonzaM: Hmm not sure if updating the user here is a good approach 🤔

t-romani: I do agree with this. 
Just followed @jiv upper recommendation, I might have understood it all wrong tho. Thoughts? @gonzaM @jiv

jiv: we talked about this in person. I suggested adding the clause only in the notify, not here. Good call @gonzaM 👍

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -8,8 +8,17 @@ def sign_in_params
...
+        return unless current_user.player_id.nil?
```
JuanfraM: it's clearer:
`return if current_user.player_id.present?`

app/services/notification_service.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def notify
+    return if matching_users.nil?
+
+    notificate_users
+  end
+
+  def set_id
+    return unless user.player_id.nil?
```
JuanfraM: it's clearer:
`return if user.player_id.present?`

app/services/notification_service.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    uri = URI.parse(ENV['ONE_SIGNAL_ID_URL'].to_s)
```
JuanfraM: isn't the env variable a string 🤔 ?

app/services/notification_service.rb
```diff
@@ -0,0 +1,74 @@
...
+    params = email_create_params(email_auth_hash)
+    response = send_request(uri, params)
+    update_player_id(response)
+  end
+
+  def notificate_users
```
JuanfraM: `notify_users` maybe? it's similar to the other method

app/services/notification_service.rb
```diff
@@ -0,0 +1,74 @@
...
+    player_ids = matching_users.map(&:player_id)
+
+    uri = URI.parse(ENV['ONE_SIGNAL_NOTIFY_URL'].to_s)
```
JuanfraM: same as above

app/services/notification_service.rb
```diff
@@ -0,0 +1,74 @@
...
+      app_id: ENV['ONE_SIGNAL_APP_ID'],
+      identifier: user.email,
+      email_auth_hash: email_auth_hash
+    }
+  end
+
+  def update_player_id(response)
+    email_player_id = JSON.parse(response.body)['id']
+    user.player_id = email_player_id
+    user.save!
```
JuanfraM: `user.update!(player_id: email_player_id)`

app/services/notification_service.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def target_match_request_params(player_ids)
+    { 'app_id': ENV['ONE_SIGNAL_APP_ID'],
+      'email_subject': I18n.t('api.email.matched_target.subject'),
+      'email_body': I18n.t('api.email.matched_target.body'),
+      'include_player_ids': player_ids }
```
JuanfraM: align this hash, also I think that keys don't need `' '`

app/models/target.rb
```diff
@@ -30,12 +30,43 @@ class Target < ApplicationRecord
...
   def limit_number_of_targets
     return if user && user.targets.count < 10
 
-    errors.add(:base, I18n
-      .t('api.error.invalid_request.max_target_limit'))
+    errors.add(:base,
+               I18n.t('api.error.invalid_request.max_target_limit'))
+  end
+
+  def check_matching_targets
```
JuanfraM, line 50: I didn't see tests for this method and it's an important one. Maybe are in another PR?

JuanfraM (commented):
Good job!
I left some comments

app/services/notification_service.rb
t-romani: @JuanfraM  what do you mean by align?

app/models/target.rb
t-romani: You can find this tests in the next PR.

JuanfraM (approved):

PR #12, opened 2019-09-16
add conversations(w/index) and messages, w/tests

t-romani:
every time two targets from different users match, a conversation is created unless it already exists.
also, a message of 'you have a new match' is created.
conversations should have a maximum of 2 users each

### DIFF
--- a/app/controllers/api/v1/conversations_controller.rb
+++ b/app/controllers/api/v1/conversations_controller.rb
@@ -1,0 +1,17 @@
+module Api
+  module V1
+    class ConversationsController < Api::V1::ApiController
+      helper_method :user, :conversations
+
+      def index; end
+
+      def user
+        @user ||= current_user
+      end
+
+      def conversations
+        @conversations ||= user.conversations
+      end
+    end
+  end
+end
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -1,0 +1,45 @@
+# == Schema Information
+#
+# Table name: conversations
+#
+#  id         :bigint           not null, primary key
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class Conversation < ApplicationRecord
+  has_many :conversation_users, dependent: :destroy
+  has_many :users, through: :conversation_users
+  has_many :messages, dependent: :destroy
+
+  def self.create_conversations(matching_users)
+    user = matching_users.pop
+    matching_users.each do |match|
+      conversation = share_conversation(user, match)
+      if conversation.blank?
+        conversation = Conversation.create!
+        user.conversations << conversation
+        match.conversations << conversation
+      end
+      conversation.messages.create!(
+        text: I18n.t('api.messages.target_match')
+      )
+    end
+  end
+
+  def self.share_conversation(user, match)
+    user_conversations = user.conversations
+    match_conversations = match.conversations
+    return if user_conversations.empty? || match_conversations.empty?
+
+    user_conversations.each do |conv|
+      return conv if match_conversations.include?(Conversation.find(conv.id))
+    end
+
+    nil
+  end
+
+  def not_current_user(current_user)
+    users.reject { |user| user == current_user }
+  end
+end
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -1,0 +1,23 @@
+# == Schema Information
+#
+# Table name: messages
+#
+#  id              :bigint           not null, primary key
+#  text            :text             not null
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  conversation_id :bigint           not null
+#  user_id         :bigint
+#
+# Indexes
+#
+#  index_messages_on_conversation_id  (conversation_id)
+#  index_messages_on_user_id          (user_id)
+#
+
+class Message < ApplicationRecord
+  belongs_to :user, optional: true
+  belongs_to :conversation
+
+  validates :text, presence: true
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -41,6 +41,11 @@ class User < ApplicationRecord
 
   enum gender: { male: 0, female: 1, other: 2 }
 
+  has_many :targets, dependent: :destroy
+  has_many :conversation_users, dependent: :destroy
+  has_many :conversations, through: :conversation_users
+  has_many :messages, dependent: :destroy
+
   before_validation :init_uid
 
   validates :email, presence: true, uniqueness: {
@@ -48,8 +53,6 @@ class User < ApplicationRecord
   }
   validates :full_name, :gender, presence: true
   validates :gender, inclusion: { in: genders }
-
-  has_many :targets, dependent: :destroy
 
   private
 
--- a/app/views/api/v1/conversations/_info.json.jbuilder
+++ b/app/views/api/v1/conversations/_info.json.jbuilder
@@ -1,0 +1,4 @@
+json.extract! conversation, :id
+json.users conversation.not_current_user(user),
+           partial: 'api/v1/users/info',
+           as: :user
--- a/app/views/api/v1/conversations/show.json.jbuilder
+++ b/app/views/api/v1/conversations/show.json.jbuilder
@@ -1,0 +1,11 @@
+json.conversation do
+  json.id conversation.id
+  json.users conversation.not_current_user(@user),
+             partial: 'api/v1/users/info',
+             as: :user
+
+  json.messages @conversation.messages do |message|
+    json.message message, partial: 'api/v1/messages/info',
+                          as: :message
+  end
+end
--- a/db/migrate/20190912121620_create_conversations.rb
+++ b/db/migrate/20190912121620_create_conversations.rb
@@ -1,0 +1,12 @@
+class CreateConversations < ActiveRecord::Migration[5.2]
+  def change
+    create_table(:conversations, &:timestamps)
+
+    create_table :conversation_users, bulk: true, id: false do |t|
+      t.belongs_to :user, null: false, foreign_key: true
+      t.belongs_to :conversation, null: false, foreign_key: true
+    end
+
+    add_index(:conversation_users, %i[user_id conversation_id], unique: true)
+  end
+end
--- a/db/migrate/20190912123831_create_messages.rb
+++ b/db/migrate/20190912123831_create_messages.rb
@@ -1,0 +1,11 @@
+class CreateMessages < ActiveRecord::Migration[5.2]
+  def change
+    create_table :messages do |t|
+      t.references  :user
+      t.references  :conversation, null: false, index: true
+      t.text        :text, null: false
+
+      t.timestamps
+    end
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -9,3 +9,54 @@ titles = ['The Pragmatic Programmer', 'The Stupendous Origin', 'The Awakening of
 titles.each do |title|
   Topic.find_or_create_by(title: title)
 end
+
+# user = User.new
+# user.email = 'tiziana.romani@rootstrap.com'
+# user.password = 'aaaaaa'
+# user.password_confirmation = 'aaaaaa'
+# user.gender = 'female'
+# user.full_name = 'Sassss'
+# user.confirm
+# user.save
+
+# user = User.new
+# user.email = 'tiziana.romani1@rootstrap.com'
+# user.password = 'aaaaaa'
+# user.password_confirmation = 'aaaaaa'
+# user.gender = 'female'
+# user.full_name = 'Sassss'
+# user.confirm
+# user.save
+
+# user = User.new
+# user.email = 'tiziana.romani2@rootstrap.com'
+# user.password = 'aaaaaa'
+# user.password_confirmation = 'aaaaaa'
+# user.gender = 'female'
+# user.full_name = 'Sassss'
+# user.confirm
+# user.save
+
+User.first.targets.find_or_create_by(
+  topic_id: 1,
+  title: 'Fields',
+  radius: 500,
+  latitude: -34.9071161,
+  longitude: -56.2033275
+)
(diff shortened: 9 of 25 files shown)


app/controllers/api/v1/conversations_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Api
+  module V1
+    class ConversationsController < Api::V1::ApiController
+      helper_method :user, :conversations
+
+      def index
+        conversations
```
JuanfraM: unnecessary use of conversations here since you've declared it as a helper_method

app/controllers/api/v1/messages_controller.rb
```diff
@@ -0,0 +1,6 @@
+module Api
+  module V1
+    class MessagesController < Api::V1::ApiController
```
JuanfraM: why did you declare an empty controller?

app/models/conversation.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+# Table name: conversations
+#
+#  id         :bigint           not null, primary key
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class Conversation < ApplicationRecord
+  has_and_belongs_to_many :users
```
JuanfraM: we prefer user `has_many :through`
[RoR guides](https://github.com/rootstrap/tech-guides/blob/master/ruby/rails.md#models)

app/controllers/api/v1/conversations_controller.rb
t-romani: We had this discussion already on targets @JuanfraM , it doesnt work without that `conversations`

app/controllers/api/v1/messages_controller.rb
t-romani: because I was foreseeing the future 😌 
I'll delete it if no 🔮 allowed

app/controllers/api/v1/conversations_controller.rb
JuanfraM: mmm are you sure??
In the view you're calling `@conversations` , if you remove this variable in the controller and call just `conversations` ??

JuanfraM: We can check it on monday

app/models/conversation.rb
```diff
@@ -0,0 +1,41 @@
...
+#
+# Table name: conversations
+#
+#  id         :bigint           not null, primary key
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class Conversation < ApplicationRecord
+  has_many :conversation_users, dependent: :destroy
```
JuanfraM, line 11: You've already declared dependent destroy on the user model below

app/models/conversation.rb
```diff
@@ -0,0 +1,41 @@
...
+        conversation = Conversation.create!
+        user.conversations << conversation
+        match.conversations << conversation
+      end
+      conversation.messages.create!(text: 'You have a new match!')
+    end
+  end
+
+  def self.share_conversation(user, match)
+    return if user.conversations.nil? || match.conversations.nil?
```
JuanfraM: you should check if `empty?`, it's more appropriate

app/models/conversation.rb
```diff
@@ -0,0 +1,41 @@
...
+  def self.create_conversations(matching_users)
+    user = matching_users.pop
+    matching_users.each do |match|
+      conversation = share_conversation(user, match)
+      if conversation.blank?
+        conversation = Conversation.create!
+        user.conversations << conversation
+        match.conversations << conversation
+      end
+      conversation.messages.create!(text: 'You have a new match!')
```
JuanfraM: this text should be in the I18 file

app/models/conversation.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    user.conversations.each do |conv|
+      return conv if match.conversations.include?(Conversation.find(conv.id))
+    end
+
+    nil
+  end
+
+  def not_current_user(current_user)
+    users.map { |user| user if user != current_user }.compact
```
JuanfraM: compact removes nil elements, why here 🤔

app/models/user.rb
```diff
@@ -41,6 +41,11 @@ class User < ApplicationRecord
 
   enum gender: { male: 0, female: 1, other: 2 }
 
+  has_many :targets, dependent: :destroy
+  has_many :conversation_users, dependent: :destroy
+  has_many :conversations, through: :conversation_users
+  has_many :messages, dependent: :destroy
```
JuanfraM, line 47: `dependent: :destroy` is unnecessary here since you're deleting them through conversations

app/views/api/v1/conversations/show.json.jbuilder
```diff
@@ -0,0 +1,11 @@
+json.conversation do
+  json.id conversation.id
+  json.user conversation.not_current_user(@user),
```
JuanfraM: `json.users` ??

spec/models/target_spec.rb
```diff
@@ -34,4 +34,77 @@
     it { is_expected.to belong_to(:user) }
     it { is_expected.to belong_to(:topic) }
   end
+
+  describe 'functions' do
```
JuanfraM: `methods` is the same as functions in ruby

spec/models/target_spec.rb
```diff
@@ -34,4 +34,77 @@
...
+    describe '.check_matching_targets' do
+      context 'when targets match' do
+        subject do
+          create(:target, user: user, topic: topic,
+                          latitude: target_params[:latitude],
+                          longitude: target_params[:longitude])
+        end
+
+        it 'calls for notification job' do
+          ActiveJob::Base.queue_adapter = :test
```
JuanfraM, line 58: isn't this by default in tests?

JuanfraM (commented):
Nice job.
I left some comments

app/models/conversation.rb
t-romani: Rubocop: Specify a `:dependent` option.

app/models/conversation.rb
t-romani: because map was adding nil elements.
I have changed it tho, to:
`users.reject { |user| user == current_user }`

app/models/user.rb
t-romani: RuboCop: Specify a `:dependent` option.

spec/models/target_spec.rb
t-romani: 🤷‍♀ idk,
it makes use of it [here](https://relishapp.com/rspec/rspec-rails/docs/job-specs/job-spec)
Also, if i don't put it, it checks into default queue

app/models/conversation.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    user.conversations.each do |conv|
+      return conv if match.conversations.include?(Conversation.find(conv.id))
+    end
+
+    nil
+  end
+
+  def not_current_user(current_user)
+    users.reject { |user| user == current_user }
```
leandrodzp, line 43: Didn't know the existence of `.reject`, nice to know it! This returns all the users but `current_user`, doesn't it?

app/models/message.rb
```diff
@@ -0,0 +1,23 @@
...
+#  user_id         :bigint
+#
+# Indexes
+#
+#  index_messages_on_conversation_id  (conversation_id)
+#  index_messages_on_user_id          (user_id)
+#
+
+class Message < ApplicationRecord
+  belongs_to :user, optional: true
```
leandrodzp, line 19: Why is this optional needed? Does make sense a message without a user? (Maybe yes)

db/migrate/20190912121620_create_conversations.rb
```diff
@@ -0,0 +1,12 @@
+class CreateConversations < ActiveRecord::Migration[5.2]
+  def change
+    create_table(:conversations, &:timestamps)
+
+    create_table :conversation_users, bulk: true, id: false do |t|
```
leandrodzp, line 5: I know you removed this `id: false` in futures PR

db/migrate/20190912123831_create_messages.rb
```diff
@@ -0,0 +1,11 @@
+class CreateMessages < ActiveRecord::Migration[5.2]
+  def change
+    create_table :messages do |t|
+      t.references  :user
```
leandrodzp, line 4: Same comment here, if users are not always necessary this is ok, but if not, add also `null: false`.

db/seeds.rb
```diff
@@ -9,3 +9,54 @@
 titles.each do |title|
   Topic.find_or_create_by(title: title)
 end
+
+# user = User.new
```
leandrodzp, line 13: Why these commented lines?

spec/models/conversation_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    subject(:create_conversations) do
+      Conversation.create_conversations([user, second_user])
+    end
+
+    it 'creates a conversation' do
+      expect { subject }.to change { Conversation.count }.by(1)
+    end
+
+    it 'includes each user' do
```
leandrodzp, line 26: You could check the same for `second_user`, just to check that both users were included in the conversation.

spec/models/conversation_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      subject
+      expect(Conversation.last).to eq(user.conversations.last)
+    end
+  end
+
+  describe '.share_conversation' do
+    context 'when created conversation' do
+      let!(:conversation) { Conversation.create! }
+
+      it 'returns true' do
```
leandrodzp: I think the name of the tests is not what you are checking, is it?

spec/models/conversation_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        conversation.users << user
+        conversation.users << second_user
+        expect(Conversation.share_conversation(user, second_user))
+          .to eq(conversation)
+      end
+    end
+
+    context 'when not created a conversation between the users' do
+      let!(:conversation) { create(:conversation) }
```
leandrodzp: I think this is not necesarry, is it? Since you're not using this conversation in this test.

spec/request/conversation/index_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  context 'when valid' do
+    let!(:auth_headers) { auth_user_headers }
+
+    it 'gets a successfull answer' do
+      expect(response).to be_successful
+    end
+
+    it 'returns the correct amount of conversations' do
+      expect(parsed_data['conversations'].size)
+        .to eq(2)
```
leandrodzp: Instead of hardcoding this `2`, maybe you can use the `user.conversations.count`, does this make sense? I know it's the same, but would be better I think.

spec/request/conversation/index_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      let!(:auth_headers) { {} }
+
+      it 'shows unauthorized message' do
+        expect(response).to have_http_status(:unauthorized)
+      end
+
+      it 'returns error' do
+        expect(parsed_data['errors']).to eq(
+          [
+            'You need to sign in or sign up before continuing.'
```
leandrodzp: Use internationalization (I18n), also, if you change this message you don't have to change the tests. If you're only checking if there's any error, you could check that `parsed_data['errors']` is not empty.

spec/request/target/create_spec.rb
```diff
@@ -51,9 +53,269 @@
...
+            expect(user.conversations.first.id)
+              .to eq(Conversation.first.id)
+            expect(second_user.conversations.first.id)
+              .to eq(Conversation.first.id)
+          end
+          it 'sends matched target message' do
+            subject
+            expect(Conversation.last.messages.size).to eq(1)
+            expect(Conversation.last.messages.last.text)
+              .to eq('You have a new match!')
```
leandrodzp: Same here, use I18n

leandrodzp (commented):
A few comments, good work and nice tests!

app/models/message.rb
t-romani: When there is a target match, a new message is sent in the conversation saying 'new target match' ; you can match several times and since you only have one conversation you will see as many messages as matches. This messages dont have a user.

app/models/conversation.rb
t-romani: exactly!

db/migrate/20190912121620_create_conversations.rb
t-romani: yes 😭  silly past me

db/seeds.rb
t-romani: Because I have an issue with seeding users, cant seed an user more than once, and i cant use find_or_create_by method either. So gotta comment the lines if im seeding several times, or i would get 'email already taken' error

spec/models/conversation_spec.rb
t-romani: you are right, this used to return boolean value until it was changed

spec/request/conversation/index_spec.rb
t-romani: Thats from devise,  i dont have it in my locale yml file;
should i keep it like this or just check that there are errors?

app/views/api/v1/conversations/_info.json.jbuilder
```diff
@@ -0,0 +1,4 @@
+json.extract! conversation, :id
+json.users conversation.not_current_user(@user),
```
JuanfraM: you have a helper method `user`. Same in the other views

JuanfraM (approved):
Just one minor comment.
Good job!

spec/request/conversation/index_spec.rb
leandrodzp: I would check if `parsed_data['errors']` is not empty, what do you think?

leandrodzp (approved):

PR #13, opened 2019-09-16
add creation of messages to a conversation, w/test on message model a…

t-romani:
when creating a message, the rest of the users in the conversation receive a notification on the email

### DIFF
--- a/app/controllers/api/v1/conversations/messages_controller.rb
+++ b/app/controllers/api/v1/conversations/messages_controller.rb
@@ -1,0 +1,38 @@
+module Api
+  module V1
+    module Conversations
+      class MessagesController < Api::V1::ApiController
+        helper_method :conversation, :user
+
+        def create
+          @message = MessageService.new(conversation).create_message(
+            message_params['text'], current_user
+          )
+          render 'api/v1/messages/show'
+        end
+
+        def index
+          @messages = conversation.messages
+          render 'api/v1/messages/index'
+        end
+
+        def conversation
+          @conversation ||= current_user.conversations
+                                        .find(params[:conversation_id])
+        end
+
+        def user
+          @user ||= current_user
+        end
+
+        private
+
+        def message_params
+          params.require(:message).permit(
+            :text
+          )
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -17,7 +17,7 @@ module Api
       def set_player_id
         return if current_user.player_id.present?
 
-        NotificationService.new(current_user, nil).set_id
+        NotificationService.new(current_user).set_id
       end
     end
   end
--- a/app/services/message_service.rb
+++ b/app/services/message_service.rb
@@ -1,0 +1,27 @@
+class MessageService
+  attr_accessor :conversation
+
+  def initialize(conversation)
+    @conversation = conversation
+  end
+
+  def create_message(text, user)
+    @message = conversation.messages
+                           .create!(text: text, user_id: user.id)
+    notify(text, user)
+    @message
+  end
+
+  def notify(text, user)
+    return if target_match(text)
+
+    users = conversation.not_current_user(user)
+    NotifyRequestJob.perform_later(user, users, text)
+  end
+
+  private
+
+  def target_match(text)
+    text == I18n.t('api.messages.target_match')
+  end
+end
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -1,9 +1,11 @@
 class NotificationService
-  attr_accessor :user, :matching_users
+  attr_accessor :user, :matching_users, :text
+  ONE_SIGNAL_APP_ID = ENV['ONE_SIGNAL_APP_ID']
 
-  def initialize(user, matching_users)
+  def initialize(user, matching_users = nil, text = nil)
     @user = user
     @matching_users = matching_users
+    @text = text
   end
 
   def notify
@@ -25,7 +27,11 @@ class NotificationService
     player_ids = matching_users.map(&:player_id)
 
     uri = URI.parse(ENV['ONE_SIGNAL_NOTIFY_URL'])
-    params = target_match_request_params(player_ids)
+    params = if text.present?
+               msg_request_params(player_ids, text)
+             else
+               target_match_request_params(player_ids)
+             end
     send_request(uri, params)
   end
 
@@ -42,7 +48,7 @@ class NotificationService
   def email_create_params(email_auth_hash)
     {
       device_type: 11,
-      app_id: ENV['ONE_SIGNAL_APP_ID'],
+      app_id: ONE_SIGNAL_APP_ID,
       identifier: user.email,
       email_auth_hash: email_auth_hash
     }
@@ -55,11 +61,18 @@ class NotificationService
 
   def target_match_request_params(player_ids)
     {
-      app_id: ENV['ONE_SIGNAL_APP_ID'],
+      app_id: ONE_SIGNAL_APP_ID,
       email_subject: I18n.t('api.email.matched_target.subject'),
       email_body: I18n.t('api.email.matched_target.body'),
       include_player_ids: player_ids
     }
+  end
+
+  def msg_request_params(player_ids, text)
+    { 'app_id' => ONE_SIGNAL_APP_ID,
+      'email_subject' => I18n.t('api.email.new_message.subject'),
+      'email_body' => text,
+      'include_player_ids' => player_ids }
   end
 
   def send_request(uri, params)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,7 +11,9 @@ Rails.application.routes.draw do
         resources :users, only: :update
         resources :topics, only: :index
         resources :targets, only: %i[create index destroy]
-        resources :conversations, only: :index
+        resources :conversations, only: :index do
+          resources :messages, module: :conversations, only: %i[create index]
+        end
       end
     end
   end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -10,32 +10,17 @@ titles.each do |title|
   Topic.find_or_create_by(title: title)
 end
 
-# user = User.new
-# user.email = 'tiziana.romani@rootstrap.com'
-# user.password = 'aaaaaa'
-# user.password_confirmation = 'aaaaaa'
-# user.gender = 'female'
-# user.full_name = 'Sassss'
-# user.confirm
-# user.save
+u = User.find_or_create_by(email:"tiziana.romani@rootstrap.com", gender: 'female', full_name: 'Tiziana')
+u.password = 'aaaaaa'
+u.save
 
-# user = User.new
-# user.email = 'tiziana.romani1@rootstrap.com'
-# user.password = 'aaaaaa'
-# user.password_confirmation = 'aaaaaa'
-# user.gender = 'female'
-# user.full_name = 'Sassss'
-# user.confirm
-# user.save
+u = User.find_or_create_by(email:"tiziana.romani1@rootstrap.com", gender: 'female', full_name: 'Tiziana')
+u.password = 'aaaaaa'
+u.save
 
-# user = User.new
-# user.email = 'tiziana.romani2@rootstrap.com'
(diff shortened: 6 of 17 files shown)


app/controllers/api/v1/conversations/messages_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  module V1
+    module Conversations
+      class MessagesController < Api::V1::ApiController
+        helper_method :conversation, :user
+
+        def create
+          @message = conversation.messages
+                                 .create!(message_params)
+          @message.update!(user_id: user.id)
+          @message.notify
```
leandrodzp: Shouldn't be an after create of message this `notify` method? Or do you sometimes need to create a message without notifying the user?

app/controllers/api/v1/conversations/messages_controller.rb
```diff
@@ -0,0 +1,40 @@
+module Api
+  module V1
+    module Conversations
+      class MessagesController < Api::V1::ApiController
+        helper_method :conversation, :user
+
+        def create
+          @message = conversation.messages
+                                 .create!(message_params)
+          @message.update!(user_id: user.id)
```
leandrodzp: Instead of acceding two times to the database (first for creating and then for updating the user_id), why can't you make something like:
```ruby
conversation.messages
                      .create!(message_params.merge({ user_id: user.id }))
```
Also I'm not sure if this update is needed, are you sure it is? Since you are already getting this conversation from the current_user.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -17,7 +17,7 @@ def render_create_success
       def set_player_id
         return if current_user.player_id.present?
 
-        NotificationService.new(current_user, nil).set_id
+        NotificationService.new(current_user, nil, nil).set_id
```
leandrodzp: I would do this in a different way. Why don't you use optional parameters?. 
You could update the `initialize(user, matching_users, text)` method in notification service to look like `initialize(user, matching_users = nil, text = nil)` or `initialize(user, _matching_users, _text)` if sometimes those arguments aren't needed, so you don't have to pass `nil` params when you don't need them. Take a look into Rails guides to read more about this :)

app/services/notification_service.rb
```diff
@@ -24,8 +25,12 @@ def set_id
   def notify_users
     player_ids = matching_users.map(&:player_id)
 
-    uri = URI.parse(ENV['ONE_SIGNAL_NOTIFY_URL'])
-    params = target_match_request_params(player_ids)
+    uri = URI.parse(ENV['ONE_SIGNAL_NOTIFY_URL'].to_s)
```
leandrodzp: Is the `to_s` needed? I think Figaro works only with strings (I'm not 100% sure, take a look into this).
Also I recommend you to take a look into this related to ENVs:
https://stackoverflow.com/questions/47985660/rails-difference-between-env-fetch-and-env
Sometime you want your code to work without the environment variables defined in the `application.yml` and that's fine, but I think this is not the case and helps you not to forget setting the variables also in Heroku.

spec/models/message_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  describe '#notify' do
+    context 'when message is target match' do
+      let!(:message) do
+        create(:message, conversation: conversation,
+                         user: user,
+                         text: 'You have a new match!')
+      end
+
+      it 'does not enqueue job' do
```
leandrodzp: I'm not sure what are you testing here, since the message creation is defined with a `!` (this makes the massage to be create before executing any test code)

leandrodzp (commented):
Good job! Some comments, I couldn't understand some things you were testing.

app/controllers/api/v1/conversations/messages_controller.rb
gonzaM: I agree. Since conversation is from the `current_user`, it's not necessary to update the `user_id`

app/controllers/api/v1/conversations/messages_controller.rb
gonzaM: I don't think is wrong calling the notify from the controller avoiding callbacks. When it's possible, I prefer to avoid callbacks.
Maybe you can use a `MessageService` to encapsulate the message creation and the notification.

app/controllers/api/v1/conversations/messages_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+        def user
+          @user ||= current_user
+        end
+
+        private
+
+        def message_params
+          params.require(:message).permit(
+            :text,
+            :conversation_id
```
gonzaM: Is this `converstion_id` needed?

app/models/message.rb
```diff
@@ -20,4 +20,17 @@ class Message < ApplicationRecord
   belongs_to :conversation
 
   validates :text, presence: true
+
+  def notify
+    return if target_match
+
+    users = conversation.users.reject { |each| each == user }
+    NotifyRequestJob.perform_later(user, users, text)
+  end
```
gonzaM: Maybe you can move this method to the `MessageService` too

app/models/message.rb
```diff
@@ -20,4 +20,17 @@ class Message < ApplicationRecord
...
+
+  private
+
+  def target_match
+    text == I18n.t('api.messages.target_match')
```
gonzaM: So, you're creating a message when two users have a match, right?

app/services/notification_service.rb
```diff
@@ -62,6 +67,13 @@ def target_match_request_params(player_ids)
     }
   end
 
+  def msg_request_params(player_ids, text)
+    { 'app_id' => ENV['ONE_SIGNAL_APP_ID'],
```
gonzaM: I would move this to a constant above
`ONE_SIGNAL_APP_ID = ENV['ONE_SIGNAL_APP_ID']`

config/routes.rb
```diff
@@ -11,7 +11,11 @@
         resources :users, only: :update
         resources :topics, only: :index
         resources :targets, only: %i[create index destroy]
-        resources :conversations, only: :index
+        resources :conversations, only: :index do
+          namespace 'conversations', path: '' do
+            resources :messages, only: %i[create index]
```
gonzaM: Is it necessary the namespace if you use the resource here?
If yes, can't you use `:conversations`? Is it necessary the `path: ''`?

db/seeds.rb
```diff
@@ -10,32 +10,23 @@
...
-# user.password_confirmation = 'aaaaaa'
-# user.gender = 'female'
-# user.full_name = 'Sassss'
-# user.confirm
-# user.save
+# u = User.create!(email: "tiziana.romani2@rootstrap.com",
+#   password: 'aaaaaa',
+#   gender:'female',
+# full_name: 'Tiziana')
+# u.confirm
```
gonzaM: Why the commented code?

spec/factories/message.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :message do
+    text { Faker::Lorem.sentence(word_count: 3) }
+    user
+    association :conversation
```
gonzaM: Just `conversation` should work.
You can use `association` when the name is different, for example: `association :another_name, factory: conversation`

gonzaM (commented):
Good job 👏 Just some small comments 💪

app/controllers/api/v1/conversations/messages_controller.rb
t-romani: @leandrodzp  I do sometimes need that, since when a target matches, a new message is created (regarding new match), but notification goes for target match instead.

app/controllers/api/v1/conversations/messages_controller.rb
t-romani: That `merge` refactor is done on PR #17, ill rewrite it here as well.
Also, for some reason I have no choice but to make the merge, otherwise it doesn't register the `user_id`, even when conversation is from `current_user` @gonzaM @leandrodzp

app/services/notification_service.rb
t-romani: I did not understand the last part @leandrodzp 
Don't I want the URL stored on my env variables?

spec/models/message_spec.rb
t-romani: This tests than when the function notify on the message is called, if it was a target match message, it won't notify anyone.

app/controllers/api/v1/conversations/messages_controller.rb
t-romani: it is not.

app/models/message.rb
t-romani: yes @gonzaM , with no user assigned

config/routes.rb
t-romani: actually, it isn't necessary, who knows what was i trying to do 😆

db/seeds.rb
t-romani: Because `find_or_create_by` does not work with users when you set password, due to some devise conflict. The solution was to set passwords afterwards:
```
u = User.find_or_create_by(email:"tiziana.romani1@rootstrap.com", gender: 'female', full_name: 'Tiziana')
u.password = 'aaaaaa'
u.save
``` 
I think this is done at some point in future PRs, I'll write it here anyways

config/routes.rb
t-romani: Actually, I do need the namespace, otherwise looks for the controller on api/v1.
Also, I put `path: ''` because if not, the routes would be 
`/api/v1/conversations/:conversation_id/conversations/messages`
🤔

app/services/message_service.rb
```diff
@@ -0,0 +1,29 @@
+class MessageService
+  attr_accessor :text, :conversation, :user
+
+  def initialize(text, conversation, user)
```
gonzaM: I think it's better just set the conversation here, and the `create_message` method, add the text and user:
```ruby
def initialize(conversation)
...
end

def create_message(user, text)
...
end
```

Then `MessageService.new(conversation).create_message(user, text)`

app/services/message_service.rb
```diff
@@ -0,0 +1,29 @@
...
+    @message = conversation.messages
+                           .create!(text: text, user_id: user.id)
+    notify
+    @message
+  end
+
+  def notify
+    return if target_match
+
+    users = conversation.users.reject { |each| each == user }
```
gonzaM: The conversation has only 2 users, right? Why do you have has_many?
I would move this to the conversation model and try to use a query instead of reject (reject is in memory, the query is in the database directly)

config/routes.rb
gonzaM: Can't you use `resources :messages, module: :conversations`?

spec/support/shared_context/conversation_between_two_users.rb
```diff
@@ -0,0 +1,27 @@
+RSpec.shared_context 'conversation_between_two_users', shared_context: :metadata do
+  let!(:user)           { create(:user) }
+  let!(:topic)          { create(:topic) }
+  let(:target) do
+    create(:target,
+           topic_id: topic.id,
+           user: user)
```
gonzaM: I would use only one line for this

gonzaM (approved):

leandrodzp (approved):

app/services/message_service.rb
gonzaM: An example for the comment above:
In the conversation model:
```ruby
def other_parts(user)
  users.where.not(id: user.id) #or users.without_user(user.id)
end
```
(I think you have a scope to exclude a user, you can use that, for example: 
`scope :without_user, -> (user_id) { where.not(id: user.id) }`)

Then you call: `conversation.other_parts(user)`

The difference is that `reject` is in memory and `where` query in the database (so better performance)

t-romani: It was done like this to open the possibility of multiple user conversations in the future. 
Also, in the beginning I tried to do something like first_user, second_user on conversation model but didn't end up good, I needed the through model to store some attributes like unread messages for each user.

t-romani: I couldn't come up with a query for this. Any ideas?

gonzaM: I left a comment above. Did you try that?

gonzaM (approved):

PR #16, opened 2019-09-23
paginated messages

t-romani:
page has 20 msgs max


app/controllers/api/v1/conversations/messages_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        def create
+          @message = conversation.messages
+                                 .create!(message_params)
+          @message.update!(user_id: user.id)
+          @message.notify
+          render 'api/v1/messages/show'
+        end
+
+        def index
+          @pagy, @messages = pagy(conversation.messages.order('id DESC'))
```
gonzaM: You can use `order(id: :desc)`. Also, I would move this to a scope, for example: `order_desc`

app/models/conversation.rb
```diff
@@ -0,0 +1,43 @@
...
+#  updated_at :datetime         not null
+#
+
+class Conversation < ApplicationRecord
+  has_many :conversation_users, dependent: :destroy
+  has_many :users, through: :conversation_users
+  has_many :messages, dependent: :destroy
+
+  def self.create_conversations(matching_users)
+    user = matching_users.pop
```
gonzaM: So, matching_users include the user and all the matches for the user? If yes, don't you think is better to separate them in the params, so it's clearer? `self.create_conversations(user, matching_users)`

app/models/conversation.rb
```diff
@@ -0,0 +1,43 @@
...
+        user.conversations << conversation
+        match.conversations << conversation
+      end
+      conversation.messages.create!(
+        text: I18n.t('api.messages.target_match')
+      )
+    end
+  end
+
+  def self.share_conversation(user, match)
```
gonzaM: Can't you use queries for this? A scope for example? `scope :between, -> (user_a, user_b) { where.... }`
and then you can call `Coversation.between(user, match)`
Also, using a query will have better performance than making this on memory.

app/models/conversation.rb
```diff
@@ -0,0 +1,43 @@
...
+    user.conversations.each do |conv|
+      return conv if match.conversations.include?(Conversation.find(conv.id))
+    end
+
+    nil
+  end
+
+  def not_current_user(current_user)
+    users.reject { |user| user == current_user }
+  end
```
gonzaM: You can use a scope here too.
For example: In the User class: `:not_id, -> (user_id) { where.not(id: user_id) }`. Then you can call `users.not_id(current_user)`. With this approach, you will query the database directly.

app/models/message.rb
```diff
@@ -0,0 +1,36 @@
...
+#  user_id         :bigint
+#
+# Indexes
+#
+#  index_messages_on_conversation_id  (conversation_id)
+#  index_messages_on_user_id          (user_id)
+#
+
+class Message < ApplicationRecord
+  belongs_to :user, optional: true
```
gonzaM: 🤔  is this because of the match message?

db/migrate/20190912121620_create_conversations.rb
```diff
@@ -0,0 +1,12 @@
+class CreateConversations < ActiveRecord::Migration[5.2]
+  def change
+    create_table(:conversations, &:timestamps)
+
+    create_table :conversation_users, bulk: true, id: false do |t|
```
gonzaM: id: false ?

db/migrate/20190912121620_create_conversations.rb
```diff
@@ -0,0 +1,12 @@
+class CreateConversations < ActiveRecord::Migration[5.2]
+  def change
+    create_table(:conversations, &:timestamps)
```
gonzaM: Why this?

db/seeds.rb
```diff
@@ -9,3 +9,45 @@
...
+#   password: 'aaaaaa',
+#   gender:'female',
+# full_name: 'Tiziana')
+# u.confirm
+
+# u = User.create!(email: "tiziana.romani2@rootstrap.com",
+#   password: 'aaaaaa',
+#   gender:'female',
+# full_name: 'Tiziana')
+# u.confirm
```
gonzaM: Remove commented code

spec/factories/conversation.rb
```diff
@@ -0,0 +1,4 @@
+FactoryBot.define do
+  factory :conversation do
+  end
+end
```
gonzaM: If this is empty, remove the file.

spec/factories/conversation_user.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :conversation_user do
+    association :conversation
```
gonzaM: `association` is not necessary

spec/factories/message.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :message do
+    text { Faker::Lorem.sentence(word_count: 3) }
+    user
+    association :conversation
```
gonzaM: Same as above

gonzaM (commented):
Some comments. I'm confused about the code in this PR and the code in other PRs. So I will continue reviewen this when some PRs are merged haha

app/models/conversation.rb
t-romani: Here the `not_current_user`  refers to the rest of the users on the conversation that are not current user. For now is just one but it opens up the possibility to be more in the future.
This is used mainly on conversation views, the idea is to show within the conversation, the list of users that are part of it, not showing current user. 
Maybe I can still do a scope, but it would be definitely tricky.

app/models/message.rb
t-romani: Exactly

db/migrate/20190912121620_create_conversations.rb
t-romani: I know 😢 I messed it up, I thought I wouldn't need that ID, since I could manage with (user_id, conversation_id). BUT NO
I arrange this here #17 , even ask you for forgiveness 🌝

db/migrate/20190912121620_create_conversations.rb
t-romani: I don't know what is this doing here. This is from old commits, probably messed it when rebasing. Will rebase again before upload to see if it goes away.

app/models/conversation.rb
t-romani: Will use detect instead, as we spoke earlier.

spec/factories/conversation.rb
t-romani: I need this when calling it from factories that have an association with conversation. Otherwise initializing anything that contains a conversation fails with 

>  Trait not registered: "conversation"

app/models/conversation.rb
```diff
@@ -32,11 +31,9 @@ def self.share_conversation(user, match)
     match_conversations = match.conversations
     return if user_conversations.empty? || match_conversations.empty?
 
-    user_conversations.each do |conv|
-      return conv if match_conversations.include?(Conversation.find(conv.id))
+    user_conversations.detect do |conv|
+      match_conversations.include?(Conversation.find(conv.id))
```
gonzaM: Why not `include?(conv)`?

app/models/message.rb
```diff
@@ -20,4 +20,8 @@ class Message < ApplicationRecord
   belongs_to :conversation
 
   validates :text, presence: true
+
+  scope :order_desc, lambda {
+    order(id: :desc)
+  }
```
gonzaM: You can use one line for this: `scope :order_desc, -> { order(id: :desc)  }`

spec/request/messages/index_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'rails_helper'
+
+describe 'GET #index messages', type: :request do
+  include_context 'conversation_between_two_users'
+
+  let!(:messages) do
+    create_list(:message, 21, user: user, conversation: conversation)
```
gonzaM: Wow, 21 is a lot. When possible, create fewer objects so you can decrease the time for all the tests.
Can't you modify the page number for the test?

gonzaM (commented):
Small comments

spec/request/messages/index_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  subject do
+    get api_v1_conversation_messages_path(
+      conversation_id: conversation.id,
+      page: page
+    ), headers: auth_headers,
+       as: :json
+  end
+
+  context 'when pagination every 20 messages' do
```
gonzaM: Remember to update this

spec/request/messages/index_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        expect(parsed_data['messages'].first['id']).to eq(Message.last.id)
+      end
+    end
+
+    context 'when getting second page' do
+      let!(:page) { 2 }
+
+      it 'returns 1 message' do
+        subject
+        expect(parsed_data['messages'].count).to eq(1)
```
gonzaM, line 40: If you created 2 messages, why this is 1 and not 0? Where are you creating another message?

gonzaM (approved):

spec/request/messages/index_spec.rb
t-romani: On target match, a userless message is created.

leandrodzp (approved):

PR #17, opened 2019-09-24
add unread messages on conversation index, w/tests

t-romani:
Yes @gonzaM , I did change old migrations, but this time I had no choice, I messed it big in the past creating an idless table and had to make up for it🍤 .  You learn something new every day...😄

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -12,6 +12,12 @@ class Conversation < ApplicationRecord
   has_many :users, through: :conversation_users
   has_many :messages, dependent: :destroy
 
+  scope :with_unread_messages_count_for, lambda { |user_id|
+    select('conversations.*, conversation_users.unread_messages AS unread_count')
+      .joins(:conversation_users)
+      .where(conversation_users: { user_id: user_id })
+  }
+
   def self.create_conversations(user, matching_users)
     matching_users.each do |match|
       conversation = share_conversation(user, match)
@@ -21,7 +27,8 @@ class Conversation < ApplicationRecord
         match.conversations << conversation
       end
       conversation.messages.create!(
-        text: I18n.t('api.messages.target_match')
+        text: I18n.t('api.messages.target_match'),
+        message_type: Message.message_types[:match]
       )
     end
   end
@@ -38,5 +45,19 @@ class Conversation < ApplicationRecord
 
   def not_current_user(current_user)
     users.where.not(id: current_user.id).to_a
+  end
+
+  def new_message(user_id)
+    conversations_with_new_messages = ConversationUser.not_for_user(user_id, id)
+    conversations_with_new_messages.each do |conversation_user|
+      conversation_user.update!(unread_messages:
+          (conversation_user.unread_messages + 1))
+    end
+  end
+
+  def read_messages(user_id)
+    conversation_with_read_messages =
+      conversation_users.find_by(user_id: user_id)
+    conversation_with_read_messages.update!(unread_messages: 0)
   end
 end
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -3,6 +3,7 @@
 # Table name: messages
 #
 #  id              :bigint           not null, primary key
+#  message_type    :integer          not null
 #  text            :text             not null
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
@@ -16,10 +17,22 @@
 #
 
 class Message < ApplicationRecord
+  enum message_type: { match: 0, user: 1 }
+
   belongs_to :user, optional: true
   belongs_to :conversation
 
-  validates :text, presence: true
+  validates :text, :message_type, presence: true
 
   scope :order_desc, -> { order(id: :desc) }
+
+  after_create :increment_unread
+
+  private
+
+  def increment_unread
+    return if match?
+
+    conversation.new_message(user.id)
+  end
 end
--- a/db/migrate/20190912121620_create_conversations.rb
+++ b/db/migrate/20190912121620_create_conversations.rb
@@ -2,7 +2,7 @@ class CreateConversations < ActiveRecord::Migration[5.2]
   def change
     create_table(:conversations, &:timestamps)
 
-    create_table :conversation_users, bulk: true, id: false do |t|
+    create_table :conversation_users, bulk: true do |t|
       t.belongs_to :user, null: false, foreign_key: true
       t.belongs_to :conversation, null: false, foreign_key: true
     end
--- a/db/migrate/20190924114916_add_unread_messages_to_conversation_users.rb
+++ b/db/migrate/20190924114916_add_unread_messages_to_conversation_users.rb
@@ -1,0 +1,6 @@
+class AddUnreadMessagesToConversationUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :conversation_users, :unread_messages,
+               :integer, default: 0, null: false
+  end
+end
--- a/db/migrate/20191004160239_add_enum_to_messages.rb
+++ b/db/migrate/20191004160239_add_enum_to_messages.rb
@@ -1,0 +1,6 @@
+class AddEnumToMessages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :messages, :message_type,
+               :integer, null: false, default: 0
+  end
+end
--- a/spec/factories/message.rb
+++ b/spec/factories/message.rb
@@ -1,7 +1,13 @@
 FactoryBot.define do
   factory :message do
-    text { Faker::Lorem.sentence(word_count: 3) }
+    text          { Faker::Lorem.sentence(word_count: 3) }
+    message_type  { Message.message_types[:user] }
     user
     conversation
+
+    trait :match do
+      message_type { Message.message_types[:match] }
+      text         { 'You have a new match!' }
+    end
   end
 end
--- a/spec/helpers.rb
+++ b/spec/helpers.rb
@@ -2,6 +2,15 @@ def auth_user_headers
   user.create_new_auth_token
 end
 
+def another_auth_user_headers
+  another_user.create_new_auth_token
+end
+
 def parsed_data
   JSON.parse(response.body)
+end
+
+def conversation_user
+  ConversationUser.where(conversation_id: conversation.id)
+                  .where(user_id: user.id).first
 end
--- a/spec/models/conversation_spec.rb
+++ b/spec/models/conversation_spec.rb
@@ -1,3 +1,12 @@
+# == Schema Information
+#
+# Table name: conversations
+#
+#  id         :bigint           not null, primary key
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
 require 'rails_helper'
 
 describe Conversation, type: :model do
@@ -46,6 +55,53 @@ describe Conversation, type: :model do
       it 'returns nil' do
         expect(Conversation.share_conversation(user, second_user)).to eq(nil)
       end
+    end
+  end
+
+  describe '.with_unread_messages_count_for' do
+    include_context 'conversation_between_two_users'
+
+    subject { Conversation.with_unread_messages_count_for(user.id) }
+
+    it 'returns the conversations for the user' do
+      expect(subject.size).to eq(user.conversations.size)
+    end
+
+    it 'returns 0 unread messages for the conversation' do
+      expect(subject.find(conversation.id).unread_count).to eq(0)
+    end
+  end
+
+  describe '#new_message' do
+    include_context 'conversation_between_two_users'
+
+    subject do
+      conversation.new_message(another_user.id)
+    end
+
+    it 'increments unread messages' do
+      expect { subject }.to change {
+        conversation_user.reload.unread_messages
+      }.by(1)
+    end
+  end
+
+  describe '#read_messages' do
+    include_context 'conversation_between_two_users'
+
+    subject do
+      conversation.read_messages(user.id)
+    end
+
+    before do
+      conversation.new_message(another_user.id)
+      conversation.new_message(another_user.id)
+    end
+
+    it 'sets unread messages value to 0' do
+      expect { subject }.to change {
+        conversation_user.reload.unread_messages
+      }.by(-2)
     end
   end
 end
--- a/spec/models/conversation_user_spec.rb
+++ b/spec/models/conversation_user_spec.rb
@@ -19,14 +19,16 @@
 #  fk_rails_...  (user_id => users.id)
 #
 
-class ConversationUser < ApplicationRecord
-  belongs_to :user
-  belongs_to :conversation
+require 'rails_helper'
 
-  validates :unread_messages, numericality: { only_integer: true }
+describe ConversationUser, type: :model do
+  include_context 'conversation_between_two_users'
 
-  scope :not_for_user, lambda { |user_id, conversation_id|
-    where(conversation_id: conversation_id)
-      .where.not(user_id: user_id)
-  }
+  subject { conversation_user }
+
+  describe 'validations' do
+    it { is_expected.to belong_to(:user) }
+    it { is_expected.to belong_to(:conversation) }
+    it { is_expected.to validate_numericality_of(:unread_messages) }
+  end
 end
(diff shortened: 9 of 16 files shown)


app/models/conversation_user.rb
```diff
@@ -20,4 +22,28 @@
 class ConversationUser < ApplicationRecord
   belongs_to :user
   belongs_to :conversation
+
+  validates :unread_messages, numericality: { only_integer: true }
+
+  def self.conversations_unread_msgs(user_id)
+    array = ConversationUser.where('user_id = ?', user_id)
```
leandrodzp: What about `ConversationUser.where(user_id: user_id)`, I think this looks much cleaner. WDYT?

app/models/conversation_user.rb
```diff
@@ -20,4 +22,28 @@
...
+    Hash[array.map do |conv_user|
+      [conv_user.conversation_id, conv_user.unread_messages]
+    end]
+  end
+
+  def self.new_message(conversation_id, user_id)
+    instances = ConversationUser.where('conversation_id = ?', conversation_id)
```
leandrodzp: Same here :)

leandrodzp: I think you could choose a more mnemotechnic name instead of instances.

app/models/conversation_user.rb
```diff
@@ -20,4 +22,28 @@
...
+                                .where('user_id != ?', user_id)
+    instances.each do |instance|
+      instance.update!(unread_messages:
```
leandrodzp: I think you could remove this logic to a method in the conversation model, WDYT?
Also check the `instance` variable name.

db/migrate/20190924114916_add_unread_messages_to_conversation_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddUnreadMessagesToConversationUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :conversation_users, :unread_messages,
+               :Integer, default: 0, null: false
```
leandrodzp: Oops `integer` should have lower case here

spec/models/conversation_user_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      expect(subject[conversation.id]).to eq(0)
+    end
+  end
+
+  describe '.new_message' do
+    subject do
+      ConversationUser.new_message(conversation.id, another_user.id)
+    end
+
+    it 'increments unread messages' do
```
leandrodzp: What about using `expect{subject}.to change{}.from(0).to(1)...`? Check the proper syntax here:
https://www.rubydoc.info/gems/rspec-expectations/RSpec%2FMatchers:change

spec/models/conversation_user_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    subject do
+      ConversationUser.read_messages(conversation.id, user.id)
+    end
+
+    before do
+      ConversationUser.new_message(conversation.id, another_user.id)
+      ConversationUser.new_message(conversation.id, another_user.id)
+    end
+
+    it 'sets unread messages value to 0' do
```
leandrodzp: Same here :)

spec/factories/conversation_user.rb
```diff
@@ -2,5 +2,6 @@
   factory :conversation_user do
     association :conversation
     association :user
+    unread_messages { 0 }
```
leandrodzp: This is not necessary since the default value of `unread_messages` is `0` :)

spec/helpers.rb
```diff
@@ -2,6 +2,15 @@ def auth_user_headers
...
 
+def another_auth_user_headers
+  another_user.create_new_auth_token
+end
+
 def parsed_data
   JSON.parse(response.body)
 end
+
+def conversation_user
```
leandrodzp, line 13: Maybe you can create a scope in the model, just a suggestion to avoid this. Also it could be helpful in a future.

app/models/message.rb
```diff
@@ -21,16 +21,25 @@ class Message < ApplicationRecord
 
   validates :text, presence: true
 
-  def notify
+  def after_creation
```
leandrodzp: I would use a callback `after_create :notify, :increment_unread` and remove this method.

leandrodzp (commented):
Great job, just small comments 🕺

app/models/message.rb
t-romani: I tried it like this at first, but was not viable since it requires to have the user set first, which is done after creation, by the controller... unless there is a way in which I can set both conversation and user ids at the same time, but I couldnt find any,  going through  user.conversations.messages.create! doesnt work. 
@leandrodzp cant tag 😢

app/models/conversation_user.rb
t-romani: Do you want me to move 
`new_message`, and `read_messages` logic to the conversation model? 
If that is the case, I cant agree, since both functions are modifying ConversationUser instances, instead of Conversation ones. 
@leandrodzp cant tag 😞

leandrodzp: No, I'm talking about
```ruby 
conversation_user.update!(unread_messages:
          (conversation_user.unread_messages + 1))
```
You can make a method like `incremet_unread(number)`, WDYT?

leandrodzp: As we talked via Slack, let's keep this this way since it makes sense with the way you implemented all the other stuff.

leandrodzp (approved):
Good job :)

app/models/conversation.rb
```diff
@@ -39,4 +39,28 @@ def self.share_conversation(user, match)
   def not_current_user(current_user)
     users.where.not(id: current_user.id).to_a
   end
+
+  def self.conversations_unread_msgs(user_id)
+    conversations = ConversationUser.where(user_id: user_id)
+    Hash[conversations.map do |conv_user|
+      [conv_user.conversation_id, conv_user.unread_messages]
+    end]
```
gonzaM: I think you can use something like:
```ruby

Conversation
  .select('conversations.id, conversation_users.unread_messages AS unread_count')
  .joins(:conversation_users)
  .where(conversation_users: { user_id: user_id })
```

It's better to return an ActiveRecordRelation than a hash because you can manipulate it better.

gonzaM: Also, I think it's not necessary the `conversations_` because you're in the `Conversation` model.
I would rename as: `self.with_unread_messages_count_for(user_id)` then you call `Conversation.with_unread_messages_count_for(user.id)`

app/models/conversation.rb
```diff
@@ -39,4 +39,28 @@ def self.share_conversation(user, match)
...
+  end
+
+  def new_message(user_id)
+    conversations_with_new_messages =
+      ConversationUser.where(conversation_id: id)
+                      .where.not(user_id: user_id)
+    conversations_with_new_messages.each do |conversation|
+      conversation.update!(unread_messages:
```
gonzaM: You can use `increment!`. Check here: https://apidock.com/rails/ActiveRecord/Base/increment!

app/models/conversation.rb
```diff
@@ -39,4 +39,28 @@ def self.share_conversation(user, match)
...
+    Hash[conversations.map do |conv_user|
+      [conv_user.conversation_id, conv_user.unread_messages]
+    end]
+  end
+
+  def new_message(user_id)
+    conversations_with_new_messages =
+      ConversationUser.where(conversation_id: id)
+                      .where.not(user_id: user_id)
+    conversations_with_new_messages.each do |conversation|
```
gonzaM: I would rename as `conversation_user` because it's not a conversation.

app/models/conversation.rb
```diff
@@ -39,4 +39,28 @@ def self.share_conversation(user, match)
...
+      conversation.update!(unread_messages:
+          (conversation.unread_messages + 1))
+    end
+  end
+
+  def read_messages(user_id)
+    conversations_with_read_messages =
+      ConversationUser.where(conversation_id: id)
```
gonzaM: Same here: `conversation_users.find_by(user_id: user_id)`

gonzaM: Also, the `find_by` will return only one conversation_user, is that the idea? If so, `conversations_with_read_messages` should not be plural.

app/models/conversation_user.rb
```diff
@@ -20,4 +22,10 @@
 class ConversationUser < ApplicationRecord
   belongs_to :user
   belongs_to :conversation
+
+  validates :unread_messages, numericality: { only_integer: true }
+
+  scope :conversation_user_instance, lambda { |params|
+    where(conversation_id: params[0]).find_by(user_id: params[1])
+  }
```
gonzaM, line 31: This should not be a scope. Scopes generally return collections, not instances.

app/models/message.rb
```diff
@@ -22,4 +22,18 @@ class Message < ApplicationRecord
   validates :text, presence: true
 
   scope :order_desc, -> { order(id: :desc) }
+
+  after_create :increment_unread
+
+  private
+
+  def target_match
```
gonzaM: What do you think about adding an `enum` to know if a message is a text or is a match?
With the current approach, if the text changes, you will not be able to identify the matches.
Also, the `enum` will provide you with methods like `match?` to know if the message is a match.
WDYT?
Let me know if you want me to explain better the benefits of an enum.

db/migrate/20190912121620_create_conversations.rb
```diff
@@ -2,7 +2,7 @@ class CreateConversations < ActiveRecord::Migration[5.2]
   def change
     create_table(:conversations, &:timestamps)
 
-    create_table :conversation_users, bulk: true, id: false do |t|
+    create_table :conversation_users, bulk: true do |t|
```
gonzaM, line 5: You can add a migration to remove the table and create it again if necessary. Instead of modifying the existing migration.

spec/helpers.rb
gonzaM, line 13: Why this helper? 🤔

app/models/conversation.rb
t-romani: Will not change it, since, as we spoke: 
Avoid using `increment!` because it skips validations. (convention:Rails/SkipsModelValidations)

db/migrate/20190912121620_create_conversations.rb
t-romani: Cool, for next time 😜

spec/helpers.rb
t-romani: To refer to the conversation_user instance im currrently using. 
This is used in conversation spec to check the unread_messages on message related methods.

gonzaM (approved):

app/models/conversation.rb
```diff
@@ -21,7 +21,8 @@ def self.create_conversations(user, matching_users)
         match.conversations << conversation
       end
       conversation.messages.create!(
-        text: I18n.t('api.messages.target_match')
+        text: I18n.t('api.messages.target_match'),
+        message_type: 'match'
```
gonzaM: It's better if you use: `Message.message_types[:match]` so it's more dynamic.

app/models/conversation.rb
```diff
@@ -39,4 +40,26 @@ def self.share_conversation(user, match)
   def not_current_user(current_user)
     users.where.not(id: current_user.id).to_a
   end
+
+  def self.with_unread_messages_count_for(user_id)
+    Conversation
```
gonzaM: I would move this to scope.
Also, when using a method like this one, it's better if you use `sefl.class` so if the name changes, you don't need to change here.

app/models/conversation.rb
```diff
@@ -39,4 +40,26 @@ def self.share_conversation(user, match)
...
+      .select('conversations.*, conversation_users.unread_messages AS unread_count')
+      .joins(:conversation_users)
+      .where(conversation_users: { user_id: user_id })
+  end
+
+  def new_message(user_id)
+    conversations_with_new_messages =
+      conversation_users.where.not(user_id: user_id)
```
gonzaM: I would move this to a scope `not_for_user(user_id)` so you can reuse if necessary.

app/models/conversation.rb
```diff
@@ -39,4 +40,26 @@ def self.share_conversation(user, match)
...
+    conversations_with_new_messages.each do |conversation_user|
+      conversation_user.update!(unread_messages:
```
gonzaM, line 53: Since this is logic for the conversation_user, I would move this to that model. For example a method `increment_messages`.
Then you can call: `conversations_with_new_messages.each(&:increment_messages)`

app/models/conversation.rb
```diff
@@ -39,4 +40,26 @@ def self.share_conversation(user, match)
...
+          (conversation_user.unread_messages + 1))
+    end
+  end
+
+  def read_messages(user_id)
+    conversation_with_read_messages =
+      conversation_users.find_by(user_id: user_id)
+    conversation_with_read_messages.update!(unread_messages: 0)
```
gonzaM, line 61: Same here

db/migrate/20191004160239_add_enum_to_messages.rb
```diff
@@ -0,0 +1,6 @@
+class AddEnumToMessages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :messages, :message_type,
```
gonzaM, line 3: You can add a default value so you don't need to specify the type always.

spec/factories/message.rb
```diff
@@ -1,6 +1,7 @@
 FactoryBot.define do
   factory :message do
     text { Faker::Lorem.sentence(word_count: 3) }
+    message_type { 'user' }
```
gonzaM: I would align this with the above one.

spec/helpers.rb
gonzaM: I don't think it's necessary to add this.

spec/models/conversation_spec.rb
```diff
@@ -48,4 +57,51 @@
       end
     end
   end
+
+  describe '.with_unread_messages_count_for' do
+    include_context 'conversation_between_two_users'
+
+    subject { Conversation.with_unread_messages_count_for(user.id) }
+
+    it 'returns a hash with a key for every user conversation' do
```
gonzaM: hash?

spec/models/message_spec.rb
```diff
@@ -28,5 +29,36 @@
   describe 'validations' do
     it { is_expected.to validate_presence_of(:text) }
     it { is_expected.to belong_to(:conversation) }
+    it { is_expected.to validate_presence_of(:message_type) }
+  end
+
+  describe '#increment_unread' do
+    context 'when message is target match' do
+      let!(:message) do
+        create(:message, conversation: conversation,
```
gonzaM: You can create a trait. so you can call `create(:message, :match, ...)` and you don't need to specify `message_type: 'match'` every time.

spec/models/message_spec.rb
```diff
@@ -28,5 +29,36 @@
...
+        message.send(:increment_unread)
+      end
+    end
+
+    context 'when message is from user' do
+      let!(:message) do
+        create(:message,
+               conversation: conversation,
+               user: user,
+               message_type: 'user')
```
gonzaM: Same here

spec/request/messages/create_spec.rb
```diff
@@ -34,5 +34,17 @@
...
+    context 'when unread messages on the other user conversation' do
+      it 'returns correct amount of unread messages on conversation' do
+        get api_v1_conversations_path, headers: another_auth_user_headers, as: :json
+        expect(parsed_data['conversations'].first['unread_messages'])
+          .to eq(0)
+        subject
+        get api_v1_conversations_path, headers: another_auth_user_headers, as: :json
+        expect(parsed_data['conversations'].first['unread_messages'])
+          .to eq(1)
+      end
```
gonzaM, line 47: It's not a good idea to call the endpoint twice in the same it. If you need to set a state before a call, use a context. Let me know if it's not clear.

gonzaM (approved):
As we talked, the PR is great, and the improvements look awesome.
I left some comments just to let you know some small things that you can improve.
Feel free to merge the PR when you need it 💪.

app/models/conversation.rb
t-romani: I had done it like this in the past. But then someone suggested I should do it on conversation side since the methods were proper from a conversation. I don't think ill change it back. But for next time I know :)

PR #19, opened 2019-09-25
clean targets task

t-romani:
deletes created targets after 1 week

TODO
heroku setup, scheduler..


lib/tasks/clean_targets.rake
```diff
@@ -0,0 +1,5 @@
+task clean_targets: :environment do
+  Target.find_each do |target|
+    target.destroy! if (Date.current - target.created_at.to_date) >= 7
```
gonzaM: You can use something like: `...if  target.created_at < 7.days.ago`

gonzaM (approved):

Alxzu (approved):

JuanfraM (approved):

PR #21, opened 2019-09-25
activeadmin configurations

t-romani:
kinda big aye? Mostly activeadmin configuration files tho😒 

avaiable on admin page:
login,
index user,
target actions,
edit information.

New model information ,
is meant to have the about_section information, and so


app/admin/information.rb
```diff
@@ -0,0 +1,8 @@
+ActiveAdmin.register Information do
+  actions :all
```
gonzaM: Is it necessary to specify this? Isn't the default all?

app/admin/targets.rb
```diff
@@ -0,0 +1,4 @@
+ActiveAdmin.register Target do
+  actions :all, except: %i[update destroy create new]
```
gonzaM: Maybe you can specify the ones that you want, show and index?

app/admin/topics.rb
```diff
@@ -0,0 +1,17 @@
...
+  #
+  # permit_params :title
+  #
+  # or
+  #
+  # permit_params do
+  #   permitted = [:title]
+  #   permitted << :other if params[:action] == 'create' && current_user.admin?
+  #   permitted
+  # end
```
gonzaM: Why the comment code?
Also, is better if you specify which attributes you want to show so you can have more control.

app/admin/users.rb
```diff
@@ -0,0 +1,3 @@
+ActiveAdmin.register User do
+  actions :all, except: %i[update destroy create new edit]
```
gonzaM: Same as above

app/controllers/api/v1/informations_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Api
+  module V1
+    class InformationsController < Api::V1::ApiController
+      skip_before_action :authenticate_user!
+
+      def show
+        @information = Information.find_by(section_title: params[:section_title])
```
gonzaM: This is not related to active admin, another PR?

app/views/api/v1/informations/_info.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.extract! information, :text
```
gonzaM: You should be consistent. You're not always using extract!. Use extract! always or not, but try to use the same approach in all the project.

db/migrate/20190925152602_create_active_admin_comments.rb
```diff
@@ -0,0 +1,16 @@
+class CreateActiveAdminComments < ActiveRecord::Migration[5.2]
+  def self.up
+    create_table :active_admin_comments do |t|
```
gonzaM, line 3: Are these necessary?

db/schema.rb
```diff
@@ -36,6 +50,32 @@
...
+    t.string "reset_password_token"
+    t.datetime "reset_password_sent_at"
+    t.datetime "remember_created_at"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["email"], name: "index_admin_users_on_email", unique: true
+    t.index ["reset_password_token"], name: "index_admin_users_on_reset_password_token", unique: true
+  end
+
+  create_table "conversation_users", force: :cascade do |t|
```
gonzaM, line 65: Coversations should not be here, right?

db/schema.rb
```diff
@@ -51,6 +91,22 @@
     t.index ["priority", "run_at"], name: "delayed_jobs_priority"
   end
 
+  create_table "information", force: :cascade do |t|
+    t.string "section_title", null: false
+    t.text "text"
+    t.index ["section_title"], name: "index_information_on_section_title", unique: true
+  end
+
+  create_table "messages", force: :cascade do |t|
```
gonzaM: Same as above

db/seeds.rb
```diff
@@ -9,3 +9,5 @@
 titles.each do |title|
   Topic.find_or_create_by(title: title)
 end
+
+AdminUser.create!(email: 'admin@example.com', password: 'password', password_confirmation: 'password') if Rails.env.development?
```
gonzaM: Maybe `find_or_create_by`?

spec/factories/admin_users.rb
```diff
@@ -0,0 +1,23 @@
...
+# Indexes
+#
+#  index_admin_users_on_email                 (email) UNIQUE
+#  index_admin_users_on_reset_password_token  (reset_password_token) UNIQUE
+#
+
+FactoryBot.define do
+  factory :admin_user do
+  end
+end
```
gonzaM: Remove the factory if you're not using it

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+# Indexes
+#
+#  index_admin_users_on_email                 (email) UNIQUE
+#  index_admin_users_on_reset_password_token  (reset_password_token) UNIQUE
+#
+
+require 'rails_helper'
+
+RSpec.describe AdminUser, type: :model do
```
gonzaM: `RSpec` is not necessary

spec/request/information/show_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+describe 'GET #show information', type: :request do
+  let!(:information) { create(:information) }
+  subject do
```
gonzaM: I would add an empty line here

spec/request/information/show_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    get api_v1_information_path(
+      section_title: information.section_title
+    ), as: :json
+  end
+
+  context 'when getting information' do
```
gonzaM: I don't think this context is necessary. WDYT?

db/migrate/20190925152602_create_active_admin_comments.rb
t-romani: These where added by the gem on install, so I just left those, should I delete them @gonzaM ?

db/schema.rb
t-romani: I think it might be just changing the table order in schema?

db/migrate/20190925152602_create_active_admin_comments.rb
gonzaM: I see. If you want, keep it, so you don't need to update the database, but you will no be using it.

db/schema.rb
gonzaM: Maybe the comment before was because of the the other PR that was already merged. Seems that it's okay now 👍

Alxzu (approved):
Looks good, check why are you having issues merging this (`schema.rb`)

jiv (approved):

PR #22, opened 2019-09-27
add avatar to user

t-romani:
user can add an avatar to its profile.
when no avatar is explicitly included, default one will be chosen.

### DIFF
--- a/app/views/api/v1/users/_info.json.jbuilder
+++ b/app/views/api/v1/users/_info.json.jbuilder
@@ -1,3 +1,4 @@
-json.id       user.id
-json.name     user.full_name
-json.email    user.email
+json.id        user.id
+json.name      user.full_name
+json.email     user.email
+json.avatar    polymorphic_url(user.avatar)
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -47,4 +47,20 @@ describe User, type: :model do
         .scoped_to(:provider).ignoring_case_sensitivity
     }
   end
+
+  context 'when image is present' do
+    let!(:user) { create(:user, :with_avatar) }
+
+    it 'is valid' do
+      expect(User.last.avatar.attached?).to eq(true)
+    end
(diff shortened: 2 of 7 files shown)


gonzaM (approved):

Alxzu:
I think either the description or the feature on this PR wasn't complete. If the feature is `Add avatar to a User` then you should be testing that, not just the empty case. Otherwise, the description of this PR is too short 😛

spec/models/user_spec.rb
```diff
@@ -47,4 +47,20 @@
...
+
+    it 'is valid' do
+      expect(User.last.avatar.attached?).to eq(true)
+    end
+  end
+
+  context 'when image is nil' do
+    let!(:user) { create(:user, avatar: nil) }
+
+    it 'has default image' do
```
JuanfraM: test should said `set a default image`

JuanfraM (commented):
I think the avatar should be returned in the view of the user

app/views/api/v1/users/_info.json.jbuilder
```diff
@@ -1,3 +1,4 @@
-json.id       user.id
-json.name     user.full_name
-json.email    user.email
+json.id        user.id
+json.name      user.full_name
+json.email     user.email
+json.avatar_id user.avatar.record.id
```
JuanfraM: I think you should use `polymorphic_url ` in order to return the path of the blob created

app/views/api/v1/users/_info.json.jbuilder
```diff
@@ -1,3 +1,4 @@
-json.id       user.id
-json.name     user.full_name
-json.email    user.email
+json.id        user.id
+json.name      user.full_name
+json.email     user.email
+json.avatar polymorphic_url(user.avatar)
```
JuanfraM: align params here

JuanfraM (approved):

PR #23, opened 2019-09-30
add contact form, mail to admins on contact, contact on admin page

t-romani:
add model for contacting admins, when a contact form is created, admins receive an email.
Admins can check contact instances on the admin page.

### DIFF
--- a/app/admin/contacts.rb
+++ b/app/admin/contacts.rb
@@ -1,0 +1,3 @@
+ActiveAdmin.register Contact do
+  actions :all, except: %i[update create new edit]
+end
--- a/app/controllers/api/v1/contacts_controller.rb
+++ b/app/controllers/api/v1/contacts_controller.rb
@@ -1,0 +1,28 @@
+module Api
+  module V1
+    class ContactsController < Api::V1::ApiController
+      helper_method :user, :admins
+
+      def create
+        contact = user.contacts.create!(contact_params)
+        AdminMailer.with(user: @user, contact: contact)
+                   .contact.deliver_later
+      end
+
+      def user
+        @user ||= current_user
+      end
+
+      def admins
+        @admins ||= AdminUser.all
+      end
+
+      def contact_params
+        params.require(:contact).permit(
+          :text,
+          :user_id
+        )
+      end
+    end
+  end
+end
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -69,10 +69,12 @@ class NotificationService
   end
 
   def msg_request_params(player_ids, text)
-    { 'app_id' => ONE_SIGNAL_APP_ID,
-      'email_subject' => I18n.t('api.email.new_message.subject'),
-      'email_body' => text,
-      'include_player_ids' => player_ids }
+    {
+      app_id: ENV['ONE_SIGNAL_APP_ID'],
+      email_subject: I18n.t('api.email.new_message.subject'),
+      email_body: text,
+      include_player_ids: player_ids
+    }
   end
 
   def send_request(uri, params)
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -3,33 +3,11 @@ titles.each do |title|
   Topic.find_or_create_by(title: title)
 end
 
-User.create!(email:"tiziana.romani@rootstrap.com", gender: 'female', full_name: 'Tiziana', password: 'aaaaaa', password_confirmation: 'aaaaaa') if User.where(email: 'tiziana.romani@rootstrap.com').blank?
-User.create!(email:"tiziana.romani1@rootstrap.com", gender: 'female', full_name: 'Tiziana', password: 'aaaaaa', password_confirmation: 'aaaaaa') if User.where(email: 'tiziana.romani1@rootstrap.com').blank?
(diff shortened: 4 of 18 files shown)


app/services/notification_service.rb
```diff
@@ -69,10 +77,17 @@ def target_match_request_params(player_ids)
   end
 
   def msg_request_params(player_ids, text)
-    { 'app_id' => ONE_SIGNAL_APP_ID,
-      'email_subject' => I18n.t('api.email.new_message.subject'),
-      'email_body' => text,
-      'include_player_ids' => player_ids }
+    { app_id: ENV['ONE_SIGNAL_APP_ID'],
```
JuanfraM: these params should be aligned. Same as above

app/controllers/api/v1/contacts_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Api
+  module V1
+    class ContactsController < Api::V1::ApiController
+      helper_method :user, :admins
+
+      def create
+        Contact.create!(contact_params.merge(user_id: user.id))
+        NotificationService.new(user, admins, contact_params['text'])
```
JuanfraM: this service can be called after a new `contact` is created

JuanfraM (commented):
The PR's description said that an email is sent but you are sending a notification instead 🤔

t-romani:
@JuanfraM mere coincidence with the fact that notifications are sent by email as well 🙄 😆 

Edit: did this again using ActionMailer.

db/migrate/20190930121136_add_player_id_to_admin_users_and_contact.rb
```diff
@@ -0,0 +1,10 @@
+class AddPlayerIdToAdminUsersAndContact < ActiveRecord::Migration[5.2]
+  def change
+    add_column :admin_users, :player_id, :string
+
+    create_table :contacts do |t|
+      t.references :user, null: false
+      t.text       :text, null: false
```
JuanfraM: This should be split into two files.
One migration for creating the table and the other for adding the column

db/migrate/20191015115430_add_contact.rb
```diff
@@ -0,0 +1,8 @@
+class AddContact < ActiveRecord::Migration[5.2]
```
JuanfraM: I think that the name `CreateContact` is better, then in the files, you can search easier if you want to know which are the files to create tables and the others to modify schema.
`CreateContact`
`RemoveContact` or `DeleteContact`
`AddTextToContact`,
etc
Are different names for different actions

JuanfraM (commented):
Good job!

JuanfraM (approved):

app/admin/contacts.rb
```diff
@@ -0,0 +1,3 @@
+ActiveAdmin.register Contact do
+  actions :all, except: %i[update create new edit]
+end
```
gonzaM, line 3: For the future, it's better if you specify the things you want in the admin page. This way, you can have more control.
For example, if a new "secret" attribute is added, it will be automatically added to the admin page, but maybe you don't want that attribute in the admin dashboard.

app/controllers/api/v1/contacts_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Api
+  module V1
+    class ContactsController < Api::V1::ApiController
+      helper_method :user, :admins
+
+      def create
+        Contact.create!(contact_params.merge(user_id: user.id))
```
gonzaM: You can add in the user model `has_many :contacts`, and then here `user.contacts.create!(...)` so it's more object-oriented.

jiv (approved):

app/controllers/api/v1/contacts_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+        AdminMailer.with(user: @user, contact: Contact.last)
```
gonzaM: `create!` will return the object, so you can use that object here, right?
```ruby
contact = create!(...)
... contact: contact)
```

db/seeds.rb
```diff
@@ -3,33 +3,11 @@
...
-  topic_id: 1,
-  title: 'Mountains',
-  radius: 400,
-  latitude: -34.9071167,
-  longitude: -56.2033270
-)
-
-AdminUser.create!(email: 'admin@example.com', password: 'password', password_confirmation: 'password') if Rails.env.development? && AdminUser.where(email: 'admin@example.com').blank?
+AdminUser.create!(email: 'admin@example.com', password: 'password', password_confirmation: 'password') if Rails.env.development?
+User.create!(email: 'tiziana.romani@rootstrap.com', password: 'aaaaaa', password_confirmation: 'aaaaaa', gender: 'female', full_name:'Tiziana')
```
gonzaM: For the seed, I think it's better to use just `create` so if you run the seed again, it will not fail

gonzaM (approved):

app/controllers/api/v1/contacts_controller.rb
t-romani: I was wondering about this, since I wont ever be using contact from the user itself, I didnt think it was necessary to specify` has_many :contacts` on user model. But also I wasn't sure I could do this, implying not connecting bidirectionally the relation between the two models.  Is this last thing a must? @gonzaM

PR #24, opened 2019-10-04
add information model, controller and admin resource, w/tests

t-romani:
information as about_section for example

### DIFF
--- a/app/models/information.rb
+++ b/app/models/information.rb
@@ -1,0 +1,23 @@
+# == Schema Information
+#
+# Table name: information
+#
+#  id    :bigint           not null, primary key
+#  text  :text
+#  title :integer          not null
+#
+# Indexes
+#
+#  index_information_on_title  (title) UNIQUE
+#
+
+class Information < ApplicationRecord
+  enum title: { about: 0 }
+
+  validates :title, presence: true, uniqueness: true
+  validates :text, presence: true
+
+  def to_param
+    title
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,6 +15,7 @@ Rails.application.routes.draw do
           resources :messages, module: :conversations, only: %i[create index]
         end
         resources :contacts, only: :create, path: '/contact'
+        resources :informations, param: :title, only: :show, path: ''
       end
     end
   end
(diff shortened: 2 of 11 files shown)


app/models/information.rb
```diff
@@ -0,0 +1,21 @@
...
+#  section_title :string           not null
+#  text          :text
+#
+# Indexes
+#
+#  index_information_on_section_title  (section_title) UNIQUE
+#
+
+class Information < ApplicationRecord
+  validates :section_title, presence: true, uniqueness: true
```
JuanfraM: `section_title` can be just title. Also, an enum could be better 🤔 instead of a string , I think title can be `about`, `phone_number`, `address`, etc, right?

JuanfraM (approved):

jiv (approved):

app/models/information.rb
```diff
@@ -0,0 +1,23 @@
...
+#  text  :text
+#  title :integer          not null
+#
+# Indexes
+#
+#  index_information_on_title  (title) UNIQUE
+#
+
+class Information < ApplicationRecord
+  enum title: { about: 0 }
```
gonzaM, line 15: It's weird that you have an enum but only one option 🤔

config/routes.rb
```diff
@@ -14,6 +14,7 @@
         resources :conversations, only: :index do
           resources :messages, module: :conversations, only: %i[create index]
         end
+        resources :informations, param: :title, only: :show, path: ''
```
gonzaM, line 18: Why is `path: ''` necessary?

db/migrate/20191004134102_add_information.rb
```diff
@@ -0,0 +1,10 @@
+class AddInformation < ActiveRecord::Migration[5.2]
+  def change
+    create_table :information do |t|
+      t.integer  :title, null: false
+      t.text    :text
```
gonzaM: weird indentation

gonzaM (approved):

app/models/information.rb
t-romani: Yeah, this is supposed to grow when new information sections are included. This was @JuanfraM idea

config/routes.rb
t-romani: So the route will be: `/about_section` for example, instead of `/informations/about_section`

app/models/information.rb
JuanfraM: @t-romani  @gonzaM the idea here was to have control about the differents possible titles instead of having just a string. Later title could be:
`enum title: { about: 0, phone_number: 1, address: 2 }`
Just that.

PR #25, opened 2019-10-07
add message history to admin

t-romani:
Admin can expand a conversation seeing belonging users and the message history on it

Screen Shot 2019-10-10 at 11 18 04 AM


JuanfraM (approved):
It's ok for me.
It's a good practice to add a short description, even if the title is enough explicit. 
Also, for these cases, a screenshot of the admin view helps a lot

JuanfraM (approved):

jiv (approved):

PR #26, opened 2019-10-09
add facebook login

t-romani:
when redirecting from facebook, should be done towards route '/api/v1/users/facebook' including access-token in params
this is processed by sessions controller
if no user existed before, a new one will be created, otherwise new session params will be created.


app/controllers/api/v1/api_controller.rb
```diff
@@ -23,6 +23,8 @@ def render_not_found_response
         },
                status: :not_found
       end
+
+      Koala.config.api_version = 'v2.0'
```
JuanfraM: can you move this configuration inside the initializer folder?

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -3,6 +3,23 @@ module V1
     class SessionsController < DeviseTokenAuth::SessionsController
       include ActAsApi
 
+      def facebook
+        user_params = FacebookService.new(params[:access_token]).profile
+        @resource = User.from_social_provider 'facebook', user_params
+        custom_sign_in
+      rescue Koala::Facebook::AuthenticationError
+        render json: { error: 'Not Authorized' }, status: :forbidden
```
JuanfraM: this error should be in the locale files. Same as above

app/models/user.rb
```diff
@@ -54,6 +54,17 @@ class User < ApplicationRecord
   validates :full_name, :gender, presence: true
   validates :gender, inclusion: { in: genders }
 
+  def self.from_social_provider(provider, user_params)
+    find_or_create_by!(provider: provider, uid: user_params['id']) do |user|
+      user.password = Devise.friendly_token[0, 20]
+      user.full_name = user_params['first_name'] + ' ' + user_params['last_name']
+      user.email = user_params['email']
+      # user.avatar = user_params['picture']
```
JuanfraM: is this commented because the PR with the avatar is still in Code Review? or why?

app/models/user.rb
```diff
@@ -54,6 +54,17 @@ class User < ApplicationRecord
   validates :full_name, :gender, presence: true
   validates :gender, inclusion: { in: genders }
 
+  def self.from_social_provider(provider, user_params)
+    find_or_create_by!(provider: provider, uid: user_params['id']) do |user|
+      user.password = Devise.friendly_token[0, 20]
+      user.full_name = user_params['first_name'] + ' ' + user_params['last_name']
```
JuanfraM: you can use interpolation:
`user.full_name = "#{user_params['first_name']} #{user_params['last_name']}"`

config/application.yml.example
```diff
@@ -16,3 +16,6 @@ ONE_SIGNAL_APP_ID:
 
 ONE_SIGNAL_NOTIFY_URL: 
 ONE_SIGNAL_ID_URL: 
+
+FACEBOOK_APP_ID:
+FACEBOOK_APP_SECRET:
```
JuanfraM, line 21: good one 👌

spec/request/user/create_spec.rb
```diff
@@ -166,4 +166,40 @@
...
+    end
+
+    it 'creates an user' do
+      expect { subject }.to change(User, :count).by(1)
+    end
+
+    it 'assigns the information properly' do
+      subject
+      user = User.last
+      expect(user.full_name).to eq 'Test test'
```
JuanfraM: you can use `include_json` here

spec/support/mocks/facebook_service.rb
```diff
@@ -0,0 +1,19 @@
+class FacebookService
+  def initialize(access_token)
+    @access_token = access_token
+  end
+
+  def profile
+    unless @access_token.present? && @access_token != 'invalid'
+      raise Koala::Facebook::AuthenticationError.new 400, 'error'
```
JuanfraM: can you use `:bad_request` instead of `400`?

spec/request/session/create_spec.rb
```diff
@@ -35,6 +35,30 @@
         expect { subject }.to(change { User.last.player_id })
       end
     end
+
+    context 'when logging in with facebook' do
+      let!(:facebook_params) { { access_token: '123456', format: :json } }
+
+      subject { post facebook_api_v1_users_path, params: facebook_params, as: :json }
```
JuanfraM, line 42: don't you have to mock the one signal request here?

app/models/user.rb
t-romani: Exactly, user avatar PR is still not merged.

spec/request/session/create_spec.rb
t-romani: yes, this is done in the outer context

JuanfraM (approved):
good job!

app/models/user.rb
t-romani: Updated this one! I think you should give it a look and tell me what do you think about it. @JuanfraM

.rubocop.yml
```diff
@@ -70,6 +70,9 @@ Rails/FilePath:
  Enabled: false
 Rails/SaveBang:
  Enabled: true
+Security/Open:
+  Exclude:
+    - app/models/user.rb
```
gonzaM, line 75: Why this?

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -5,6 +5,27 @@ class SessionsController < DeviseTokenAuth::SessionsController
 
       helper_method :user
 
+      def facebook
+        user_params = FacebookService.new(params[:access_token]).profile
+        @resource = User.from_social_provider 'facebook', user_params
```
gonzaM: Use parenthesis

app/models/user.rb
```diff
@@ -57,6 +57,21 @@ class User < ApplicationRecord
   validates :full_name, :gender, presence: true
   validates :gender, inclusion: { in: genders }
 
+  def self.from_social_provider(provider, user_params)
```
gonzaM, line 61: I think it's better to move this to a UserService or something like that.
But it's okay, you don't need to change it.

gonzaM (approved):

.rubocop.yml
t-romani: Because it wont allow me to open the facebook avatar URL in order to download the file. @gonzaM

app/models/user.rb
t-romani: I literally copied most of facebook login from [here](https://github.com/rootstrap/seatmate_api)

